PR #1, opened 2015-09-08
card205: Ability to enable/disable rules in UI

oarrietadotcms:


ggranum:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/view/components/rule-engine/rule-component.ts, line 79 \[r1\]](https://reviewable.io:443/reviews/dotcms/core-web/1#-JyrvlmF-oISQB0s7Bzs)** ([raw file](https://github.com/dotcms/core-web/blob/de01d0f18318158a686649bdfe1434f29e25444c/src/view/components/rule-engine/rule-component.ts#L79)):</sup>
Let's keep as much knowledge internal to the script as possible - instead of passing the boolean, just have a toggle method.

(Sorry, the existing code you referenced should changed too :~) )

toggleEnabled(){
...
}

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dotcms/core-web/1)

PR #47, opened 2015-12-04
Card 407 request header conditionlet

oarrietadotcms:


ggranum:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/api/system/ruleengine/conditionlets/Inputs.ts, line 33 \[r2\]](https://reviewable.io:443/reviews/dotcms/core-web/47#-K4icybqBvK1jVM-57Um:-K4icybqBvK1jVM-57Un:-828283684) ([raw file](https://github.com/dotcms/core-web/blob/3e5678136923ae3d17312015b2196fe6bdb37598/src/api/system/ruleengine/conditionlets/Inputs.ts#L33)):*
We can get rid of the extra model type by using a map, which you can define like so:

private _inputs: { [key:string]:value:string}

---

*[src/api/system/ruleengine/conditionlets/Inputs.ts, line 39 \[r2\]](https://reviewable.io:443/reviews/dotcms/core-web/47#-K4id8n95mCNtVsPNAWZ:-K4id8n95mCNtVsPNAW_:-341481063) ([raw file](https://github.com/dotcms/core-web/blob/3e5678136923ae3d17312015b2196fe6bdb37598/src/api/system/ruleengine/conditionlets/Inputs.ts#L39)):*
Which would make this line: 
this.inputs[input.id] = input.label
( the initializer becomes this._inputs = { }  )

we'll have to connect i18N up eventually. Probably have the InputService get all the input resource strings for the entire conditionlet at once and cache them somewhere.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dotcms/core-web/47)*

PR #50, opened 2015-12-10
Card 550 confirmation rules

oarrietadotcms:


src/api/rule-engine/Rule.ts
```diff
@@ -82,6 +82,24 @@ export class RuleModel extends CwModel {
     return valid
   }
 
+  isEmpty():boolean {
```
ggranum: Sorry to suddenly get all stickler on you, but such is life ;~)  Can you make the isEmpty method be single-return?

actually - create two methods and call them from isEmpty: #hasActions() and #hasConditions

But still, single return points please :~)

PR #73, opened 2016-01-15
Card 184 datetime conditionlet ui

dsilvam:


src/view/components/rule-engine/condition-types/serverside-condition/serverside-condition.ts
```diff
@@ -86,10 +102,24 @@ export class ServersideCondition {
     this._inputs = [];
   }
 
+  setVisible(value, input):void {
+    if(!value) return
```
ggranum: Single Entry Single Exit. Also, but not applicable here, always use brackets.

src/view/components/rule-engine/condition-types/serverside-condition/serverside-condition.ts
```diff
@@ -86,10 +102,24 @@ export class ServersideCondition {
...
+    if(input.name === 'comparison'){
+      let idx = this._inputs.indexOf(input)
+      let comparisonObj = input.options.filter((e)=> { return e.value == value })[0]
```
ggranum: SESE. When nesting seems crazy, that means your method is doing too much on its own - create more support methods.

ggranum: I believe you can do
   input.options.filter( e => e.value == value )[0]

dsilvam: had it like that, but you told me to add brackets, or I misunderstood :P

src/view/components/semantic/elements/input-date/demo.ts
```diff
@@ -1,8 +1,68 @@
...
+    this.type = type
+    this.value = value
+    this.disabled = disabled
+    this.icon = icon || ''
+    if(this.icon.indexOf(' ') == -1 && this.icon.length > 0){
+      this.icon = (this.icon + ' icon').trim()
+    }
+  }
+
+  validateDate(date:string) {
```
ggranum: eeeek!

Can you take a look at momentjs and take a guess at how long you think it would take you to add it for use here?  http://momentjs.com/

src/view/components/rule-engine/condition-types/serverside-condition/serverside-condition.ts
dsilvam: this is very debatable. HashMap.java equals example:
```
public final boolean equals(Object o) {
            if (o == this)
                return true;
            if (o instanceof Map.Entry) {
                Map.Entry<?,?> e = (Map.Entry<?,?>)o;
                if (Objects.equals(key, e.getKey()) &&
                    Objects.equals(value, e.getValue()))
                    return true;
            }
            return false;
        } 
```

src/view/components/semantic/elements/input-date/demo.ts
```diff
@@ -1,8 +1,68 @@
...
+
+  validateTime(time:string) {
+    var time_regex = /^(10|11|12|[1-9]):[0-5][0-9]$/;
+    if (!time_regex.test(time)) {
+      throw new Error("Insert a valid time HH:MM");
+    }
+  }
+
+  validateDateTime(dateTime:string) {
+    // TODO: better match this regex for MM/DD/YYYY HH:MM
```
ggranum: -_-

Why not
```
try{
 new Date(dateTime) 
} catch(e){...}
```
???

But yeah. MomentJS.

ggranum:
Very nice. 

I am going to merge this because delaying doing so will mean you will bear the brunt of other merges, and that's not nice, given how small my change requests are. But please add another PR with those changes :+1:

src/view/components/semantic/elements/input-date/demo.ts
dsilvam: I honestly didn't check any of that, just copied from what Freddy did. Didn't touch those validations, already existed in the previous version of the 'input-date.ts'

src/view/components/rule-engine/condition-types/serverside-condition/serverside-condition.ts
ggranum: hah. who knows. Maybe I just realized how awesome it is to not have distracting 'return' keywords in the middle of a large code block. ;~)

PR #74, opened 2016-01-15
card-588 IE11 layout fixes

fmontes:
So while I was working on make this work on IE11 I discover that we don't need parent containers to have ```display: flex```.

So if we have parent containers with ```display: block``` and inside flex containers, the parent grows correctly in IE11 and all others browsers.

### DIFF
--- a/src/view/components/rule-engine/rule-component.ts
+++ b/src/view/components/rule-engine/rule-component.ts
@@ -38,7 +38,7 @@ var rsrc = {
   selector: 'rule'
 })
 @View({
-  template: `<div flex layout="column" class="cw-rule" [class.cw-hidden]="hidden" [class.cw-disabled]="!rule.enabled">
+  template: `<div class="cw-rule" [class.cw-hidden]="hidden" [class.cw-disabled]="!rule.enabled">
   <div flex layout="row" class="cw-header" *ngIf="!hidden" (click)="toggleCollapsed()">
     <div flex="70" layout="row" layout-align="start center" class="cw-header-info" *ngIf="!hidden">
       <i flex="none" class="caret icon cw-rule-caret large" [class.right]="collapsed" [class.down]="!collapsed" aria-hidden="true"></i>
@@ -87,13 +87,13 @@ var rsrc = {
     </div>
   </div>
   <div flex class="cw-accordion-body" [class.cw-hidden]="collapsed">
-    <condition-group layout="row" *ngFor="var group of groups; var i=index"
+    <condition-group *ngFor="var group of groups; var i=index"
                      [rule]="rule"
                      [group]="group"
                      [groupIndex]="i"
                      (remove)="onConditionGroupRemove($event)"
                      (change)="onConditionGroupChange($event)"></condition-group>
-    <div flex layout="column" class="cw-action-group">
+    <div class="cw-action-group">
       <div flex class="cw-action-separator">
         {{rsrc('inputs.action.firesActions') | async}}
       </div>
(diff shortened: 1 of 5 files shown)


ggranum:
Does this reflow properly when the page is resized?  I'm not feeling highly confident in taking off the flex layout of a component that is supposed to... have a  flex layout.

fmontes:
@ggranum yes, because inside there are still flex and the parent div's are just a block 100% width.

src/view/components/rule-engine/rule-condition-group-component.ts
```diff
@@ -17,7 +17,7 @@ import {ServerSideTypeModel} from "../../../api/rule-engine/ServerSideFieldModel
   selector: 'condition-group'
 })
 @View({
-  template: `<div flex layout="column" class="cw-rule-group">
+  template: `<div class="cw-rule-group">
```
ggranum: How does this container get its layout flex-direction value?  If you inspect in chrome is that attribute set?

https://css-tricks.com/snippets/css/a-guide-to-flexbox/

src/view/components/rule-engine/rule-engine.ts
```diff
@@ -17,8 +17,8 @@ const I8N_BASE:string = 'api.sites.ruleengine'
   selector: 'cw-rule-engine'
 })
 @View({
-  template: `<div flex layout="column" class="cw-rule-engine">
-  <div flex layout="column" class="cw-header">
+  template: `<div class="cw-rule-engine">
+  <div class="cw-header">
     <div flex layout="row" layout-align="space-between center">
```
ggranum: For this case I can definitely see your argument - block parent, row-layout child, flex grandchildren.

src/view/components/rule-engine/styles/rule-engine.scss
```diff
@@ -141,6 +141,7 @@ input.ng-valid.required, input.ng-valid[required], .ui.input input.ng-valid[requ
 
   .cw-condition-row, .cw-action-row {
     margin-bottom: .5em;
+    min-height: 40px;
```
ggranum: Please use em unless there's a solid reason for using px.

src/view/components/rule-engine/rule-condition-group-component.ts
fmontes: We don't need ```flex-direction``` in this parent because his childs ```.cw-condition-group-separator``` and ```cw-conditions``` are a block div and a flex div one on top of the other, just like regular blocks. Here is a screenshot with more info: https://cloudup.com/cTEVDwerXv1

src/view/components/rule-engine/rule-component.ts
```diff
@@ -38,7 +38,7 @@ var rsrc = {
   selector: 'rule'
 })
 @View({
-  template: `<div flex layout="column" class="cw-rule" [class.cw-hidden]="hidden" [class.cw-disabled]="!rule.enabled">
+  template: `<div class="cw-rule" [class.cw-hidden]="hidden" [class.cw-disabled]="!rule.enabled">
   <div flex layout="row" class="cw-header" *ngIf="!hidden" (click)="toggleCollapsed()">
```
ggranum: We might be able to remove the 'flex' from here (but leave the 'layout="row"'). Maybe. If you're still playing it's worth a try.

ggranum:
Ok, giving this my +1, but not merging it yet because I think you'll probably/maybe find some small changes to make as you test more - but I promise I won't merge anything that will cause conflicts for you :~)

PR #125, opened 2016-04-06
Adding recently created rule to a bundle, or pushing the rule, is faling

fmontes:
- Fix push publish and add to bundle modal height
- Hide push and add to bundle options when rule id or key is not defined
- Enable OK button only when environtment or bundle is set (selected)

### DIFF
--- a/src/view/components/common/push-publish/push-publish-dialog-component.ts
+++ b/src/view/components/common/push-publish/push-publish-dialog-component.ts
@@ -11,10 +11,10 @@ import {IPublishEnvironment} from "../../../../api/services/bundle-service";
     [headerText]="'Push Publish'"
     [okText]="'Push'"
     [hidden]="hidden"
-    [okEnabled]="selectedEnvironment != null"
+    [okEnabled]="selectedEnvironmentId != null"
     [errorMessage]="errorMessage"
     width="25em"
-    height="20em"
+    height="auto"
     (ok)="doPushPublish.emit(selectedEnvironmentId)"
     (cancel)="cancel.emit()">
   <cw-input-dropdown
--- a/src/view/components/rule-engine/rule-component.ts
+++ b/src/view/components/rule-engine/rule-component.ts
@@ -61,12 +61,12 @@ var rsrc = {
     PushPublishDialogContainer],
   template: `<form [ngFormModel]="formModel" #rf="ngForm">
   <cw-add-to-bundle-dialog-container
-      [assetId]="rule.id"
+      [assetId]="rule.id || rule.key"
       [hidden]="!showAddToBundleDialog"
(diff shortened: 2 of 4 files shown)


src/view/components/common/push-publish/push-publish-dialog-component.ts
```diff
@@ -11,10 +11,10 @@ import {IPublishEnvironment} from "../../../../api/services/bundle-service";
     [headerText]="'Push Publish'"
     [okText]="'Push'"
     [hidden]="hidden"
-    [okEnabled]="selectedEnvironment != null"
+    [okEnabled]="selectedEnvironmentId != null"
     [errorMessage]="errorMessage"
     width="25em"
-    height="20em"
+    height="auto"
```
ggranum: Why auto? Auto is usually a curse in disguise. Were scroll bars missing?

src/view/components/rule-engine/rule-component.ts
```diff
@@ -61,12 +61,12 @@ var rsrc = {
     PushPublishDialogContainer],
   template: `<form [ngFormModel]="formModel" #rf="ngForm">
   <cw-add-to-bundle-dialog-container
-      [assetId]="rule.id"
+      [assetId]="rule.id || rule.key"
```
ggranum: This should just be 'rule.key'.

src/view/components/rule-engine/rule-component.ts
```diff
@@ -61,12 +61,12 @@ var rsrc = {
...
       [hidden]="!showAddToBundleDialog"
       (close)="showAddToBundleDialog = false; showMoreMenu = false"></cw-add-to-bundle-dialog-container>
   <cw-push-publish-dialog-container
       [environmentStores]="environmentStores"
-      [assetId]="rule.id"
+      [assetId]="rule.id || rule.key"
```
ggranum: same. rule.key

src/view/components/rule-engine/rule-component.ts
```diff
@@ -118,8 +118,8 @@ var rsrc = {
         </div>
       </div>
       <div class="ui vertical menu" *ngIf="showMoreMenu">
-        <a class="item" (click)="showAddToBundleDialog = true; $event.stopPropagation()">Add to bundle</a>
-        <a class="item" *ngIf="environmentStores.length > 0" (click)="showPushPublishDialog = true; $event.stopPropagation()">Push Publish</a>
+        <a class="item" *ngIf="rule.id || rule.key" (click)="showAddToBundleDialog = true; $event.stopPropagation()">Add to bundle</a>
```
ggranum: also rule.key. If rule.id is ever populated when rule.key is not, then there is a different bug we need to fix.

PR #126, opened 2016-04-06
8830 dependency enhancement adding page rule to bundle

oswaldogallango:


src/api/persistence/ApiRoot.ts
```diff
@@ -27,6 +28,13 @@ export class ApiRoot {
         this.hideFireOn = (hideFireOn === 'true' || hideFireOn === '1')
         console.log('hideFireOn set to ', this.hideFireOn)
       }
+      
+      let hideRulePushOptions = ApiRoot.parseQueryParam(query, "hideRulePushOptions");
```
ggranum: this should be parsed into a field on the apiRoot instance that is injected.

PR #129, opened 2016-04-07
872 usability issue adding rules to bundle

freddyucv:


src/view/components/common/modal-dialog/dialog-component.ts
```diff
@@ -58,6 +58,10 @@ export class ModalDialogComponent {
         e.preventDefault()
         e.stopPropagation()
         this.cancel.emit(false)
+      }else if(e.keyCode == 13){
```
ggranum: Please add a comment indicating which key '13' is. Also, formatting :~)

src/view/components/semantic/modules/dropdown/dropdown.ts
```diff
@@ -240,8 +241,13 @@ export class Dropdown implements AfterViewInit, OnDestroy, ControlValueAccessor
    */
   private _applyArrowNavFix($dropdown) {
     let $searchField = $dropdown.children('input.search')
+    let enterEvent = this.enter;
     $searchField.on('keyup', (event:any)=> {
       if (DO_NOT_SEARCH_ON_THESE_KEY_EVENTS[event.keyCode]) {
+        if (event.keyCode == 13 && enterEvent){
```
ggranum: Don't test for things you know are always false. In this case, you declared the event emitter in the constructor -- it will never be null. If it WERE to become null, we'd rather find out about it from the NPE than to wonder why this feature was silently failing.

freddyucv: but if we have the cw-input-dropdown tag like this (without the enter parameter), isn't the enter parameter be null?, I wanted this parameter to be optional

  <cw-input-dropdown
      flex
      [value]="bundleStores[0]?.id"
      allowAdditions="true"
      (click)="$event.stopPropagation()"
      (change)="setSelectedBundle($event)">

ggranum: I see your concern. So, 'enter' is an output, not an input. When you declare and set it to a value it will always stay that value. Consumers of that output are binding to it, not 'setting' the value, as is the case with @Input fields.

PR #132, opened 2016-04-11
Card870 - Add message for users permissions

fmontes:


src/view/components/rule-engine/rule-engine.container.ts
```diff
@@ -91,6 +93,8 @@ export class RuleEngineContainer {
   rules$:EventEmitter<RuleModel[]> = new EventEmitter()
   ruleActions$:EventEmitter<ActionModel[]> = new EventEmitter()
   conditionGroups$:EventEmitter<ConditionGroupModel[]> = new EventEmitter()
+  globalError:string;
+  showRules:boolean
```
ggranum: Are you actually using this field? It looks like you're setting the value on the 'state' object, not 'this'.

ggranum:
I think there the two field variables on rule-engine.container.ts aren't used, but otherwise it's awesome.

jtesser:
Thanx @ggranum we are updating the pull now.

src/view/components/rule-engine/rule-engine.container.ts
fmontes: You are right, fix sent

PR #136, opened 2016-04-14
872 usability issue adding rules to bundle

freddyucv:


src/view/components/common/modal-dialog/dialog-component.ts
```diff
@@ -49,6 +50,10 @@ export class ModalDialogComponent {
       } else {
         this.removeEscapeListener()
       }
+
+      if (!change.hidden.currentValue){
```
ggranum: if(!this.hidden) ...

change.foo.currentValue will always be equal to this.foo. Not a big deal.

src/view/components/common/modal-dialog/dialog-component.ts
```diff
@@ -49,6 +50,10 @@ export class ModalDialogComponent {
...
+        setTimeout( () => this.open.emit(false), 2);
```
ggranum: Please add a comment describing why this hack is required. Because it's very much a hack and we'll need to fix it the right way eventually.

src/view/components/common/push-publish/add-to-bundle-dialog-component.ts
```diff
@@ -3,6 +3,9 @@ import {CORE_DIRECTIVES} from "angular2/common";
 import {ModalDialogComponent} from "../modal-dialog/dialog-component";
 import {Dropdown, InputOption} from "../../semantic/modules/dropdown/dropdown";
 import {IBundle} from "../../../../api/services/bundle-service";
+import {ViewChild} from "angular2/core";
+
+var $ = window['$']
```
ggranum: unused?

src/view/components/semantic/modules/dropdown/dropdown.ts
```diff
@@ -87,6 +87,10 @@ export class Dropdown implements AfterViewInit, OnDestroy, ControlValueAccessor
     this._modelValue = value;
   }
 
+  focus(){
+    this._$dropdown.children('input.search')[0].focus();
```
ggranum: You could remove the hack that delays the firing of the 'open' event by wrapping this call here in a 
```
if (!$element.exists() && this.attempts < 10){ 
   setTimeout( () => {
      this.attempts++
      this.focus()
    })
if(this.attempts >= 10 ) { // log error
} else {
  this.attempts = 0
  $element[0].focus() // try catch?
}
```

I think it would be much more clear (and much more re-usable) overall.

src/view/components/common/modal-dialog/dialog-component.ts
freddyucv: ok

src/view/components/common/push-publish/add-to-bundle-dialog-component.ts
```diff
@@ -42,9 +46,10 @@ export class AddToBundleDialogComponent {
   @Output() cancel:EventEmitter<boolean> = new EventEmitter(false)
   @Output() addToBundle:EventEmitter<IBundle> = new EventEmitter(false)
 
-  public selectedBundle:IBundle = null;
+  @ViewChild(Dropdown)
```
ggranum: Good find on using @ViewChild :~)

src/view/components/common/push-publish/add-to-bundle-dialog-component.ts
freddyucv: Yes, var $ = window['$'] is unused, I added it to test something but later I forgot remove it

PR #151, opened 2016-07-21
dotCMS/core#9322 fix issues with login and z-index layout issues

fmontes:


src/view/components/common/main-component/main-component.ts
```diff
@@ -36,7 +36,7 @@ export class MainComponent {
      * Call the logout service
      */
     logout(): void {
-        this._loginService.logOutUser().subscribe( data => {
+        this._loginService.logOutUser().subscribe(data => {
             // This line update the browser url page without reloading the page
             window.history.replaceState('index.html', 'index', '/html/ng/index.html');
```
fmontes: This is another url we need to set globally or in a config.json

PR #158, opened 2016-08-04
dotCMS/core#9397 UX: Notifications window improvements

fmontes:
Style notification items and add action links

Close notification when user clicks outside

Dismiss and mark as read notifications UI

### DIFF
--- a/src/api/pipes/lowercase-pipe.ts
+++ b/src/api/pipes/lowercase-pipe.ts
@@ -1,0 +1,19 @@
+/**
+ * Created by fmontes on 7/28/16.
+ */
+import {Pipe, PipeTransform} from '@angular/core';
+
+@Pipe({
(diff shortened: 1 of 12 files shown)


src/api/pipes/lowercase-pipe.ts
```diff
@@ -0,0 +1,19 @@
...
+    name : 'lowercase',
+})
+
+/**
+ * Put in lowercase the received string
+ */
+export class LowercasePipe implements PipeTransform {
+
+    transform(value: string, args: Array<any>): string {
+        return value === undefined || value === '' ? '' : value.toLowerCase();
```
jdotcms: I have seem things such as value === undefined || value === '' ? '' or similar, shouldn't we encapsulate them into a ValidationUtils class so can be reused around all the code.

PR #163, opened 2016-08-10
Issue 9505 problems with the rememberme functionality

freddyucv:


src/view/components/common/iframe-legacy/IframeLegacyComponent.ts
```diff
@@ -23,6 +23,7 @@ export class IframeLegacyComponent {
         this.iframe = params$.pluck<string>('id')
             .distinctUntilChanged()
             .map(id => {
+                console.log('this.menuIdUrlMatch.get( id )', this.menuIdUrlMatch.get( id ));
```
jdotcms: Shouldn't we use a library that can be turn off/on if it is on dev o prod mode

PR #171, opened 2016-08-18
dotCMS/core#9431 UI: create angular-material <select> component

fmontes:
- Add ControlValueAccessor
- Create disabled state
- Add selected state to options
- Create change event emit


src/scss/_material-design.scss
```diff
@@ -0,0 +1,30 @@
+.md-underline {
```
fmontes: I create this file to group global styling that can be use in different custom components

PR #172, opened 2016-08-22
Issue 9283 create new ng2 site selector control

freddyucv:


src/api/rule-engine/Rule.ts
```diff
@@ -299,10 +300,10 @@ export class RuleService extends CoreWebService {
     });
   }
 
-  loadRules():Observable<RuleModel[]|CwError> {
+  loadRules( site:Site ):Observable<RuleModel[]|CwError> {
```
fmontes: Why do we have to send the site? shouldn't we have this somehow globally, because send the site every time we are doing a request will sucks.

src/api/services/site-service.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.switchSiteObservable = Observable.create( observer => {
+            this.switchSiteObserver = observer;
+
+            if ( this.currentSite ) {
+                this.switchSiteObserver.next(this.currentSite);
+            }
+        });
+    }
+
+    public getAllSites(): Observable<{currentSite:Site, sites:Site[]}> {
```
fmontes: I don't understand why we don't use the ```request``` method we have in CoreWebService for this... if is not working as it is, can we update it to work in this scenario too?

src/api/services/site-service.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    private setCurrentSite(siteIdentifier:string){
+        this.currentSite = Object.assign({}, this.sites.filter( site => site.identifier === siteIdentifier)[0]);
+
+        if (this.switchSiteObserver) {
+            this.switchSiteObserver.next(this.currentSite);
+        }
+    }
+
+    switchSite(siteId:String):Observable<any> {
```
fmontes: Same thing here.

src/view/components/site-selector/dot-site-selector-component.html
```diff
@@ -0,0 +1,12 @@
...
+            (change)="switchSite($event)"
+            class="site-selector__title">
+
+    <dot-option *ngFor="let site of sites"
+                [value]="site.identifier"
+                class="site-selector__option">
+        {{site.hostName}}</dot-option>
+</dot-select>
+
+<h1 *ngIf="!sites || sites.length == 1" class="site-selector__title">{{currentSite ? currentSite.hostName : ''}}</h1>
```
fmontes: This shouldn't be here... the ```currentSite``` text will be in the ```dot-select``` we just have to style it.

src/view/components/site-selector/dot-site-selector-component.ts
```diff
@@ -0,0 +1,40 @@
...
+})
+export class SiteSelectorComponent{
+    private currentSite:string;
+    private sites:Site[];
+
+    constructor(private siteService:SiteService){
+
+    }
+
+    ngOnInit(){
```
fmontes: ```ngOnInit() {``` everywhere.

src/api/rule-engine/Rule.ts
freddyucv: We don't send the site with every request, but We need send the siteId because the end point need it.

PR #175, opened 2016-08-24
dotCMS/core#9552 Add relative date to notifications

fmontes:
- Create a custom component for dates, we could extend the component for more options
- Integrate with momentjs
- Still missing the part where we connect this with the global lang of the app


src/api/services/format-date-service.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    getLang() {
+        return moment.locale();
+    }
+
+    getRelative(time):string {
+        return moment(parseInt(time, 10)).fromNow();
+    }
+
+}
```
fmontes: I create this service so we can format all the dates as we need it... but know that I see it maybe is better to have a pipe, thoughts?

src/view/boot.ts
```diff
@@ -18,6 +18,7 @@ import {MdIconRegistry} from '@angular2-material/icon/icon';
 import {CoreWebService} from "../api/services/core-web-service";
 import {LoginService} from "../api/services/login-service";
 import {RoutingService} from "../api/services/routing-service";
+import {FormatDate} from "../api/services/format-date-service";
```
fmontes: I'm injecting this from the beginning because this service in in charge of setting the language for the date and we need to share it across the app.

src/api/services/format-date-service.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    setLang(lang:string, messages:Object):void {
+        moment.locale(lang, messages || {});
+    }
+
+    getLang() {
+        return moment.locale();
+    }
+
+    getRelative(time):string {
```
fmontes: Not sure if instead of a service we can use a pipe for formatting the date. Ideas?

src/view/boot.ts
freddyucv: not really, we just need to inject from the beginning the statefull service, and according what we talked this morning we shouldn't have many (or any at all) statefull service

src/api/services/format-date-service.ts
freddyucv: Yes, I think a pipes is a better idea, something like: { time | fromNow }

src/api/services/format-date-service.ts
```diff
@@ -0,0 +1,21 @@
+import {Injectable} from '@angular/core';
+import moment from 'moment';
+
+@Injectable()
+export class FormatDate {
+
+    constructor() {}
+
```
jdotcms: I like the setLang and getLang methods, I think they might be encapsulated into an interface, Something like LanguageAware or LanguageSupport interface.

src/api/services/format-date-service.ts
jdotcms: I might be a good idea, but should be a separated artifact in order to split it out the responsibilities.

src/view/boot.ts
jdotcms: I think it is ok meanwhile we make a research about how to keep the app state, until that part I am ok with statefull services.

fmontes: In this case, we have to wait until we have the final solution.

PR #184, opened 2016-09-07
dotCMS/core#9707 UI: Integrate i18 relative date massages changes into the notifications UI

fmontes:


src/api/services/format-date-service.ts
```diff
@@ -4,19 +4,23 @@ import moment from 'moment';
 @Injectable()
 // TODO: probably a good idea to create a interface for this, will revisit this when we defined the architecture for the lang
 export class FormatDate {
+    private m;
```
jdotcms: Question: shouldn't init it on the constructor or at least check in the getLang if this.m is undefined?

src/api/services/login-service.ts
```diff
@@ -179,6 +180,14 @@ export class LoginService extends CoreWebService {
         }).map(response => {
             this.setLogInUser(response.entity);
             this.dotcmsConfig.setUser(response.entity);
+
+            // TODO: we need to refactor the appconfig because it's not updating after login/logout
+            this.appConfigurationService.getConfigProperties().subscribe(res => {
+                let langCode = res.dotcmsConfig.configParams.user.languageId.split('_')[0];
```
jdotcms: Note: this should be returned in a more semantic way from the BE, so FE should just do user.language.code, instead of splitting the strings

PR #186, opened 2016-09-12
dotcma/core#9721 dotconfig user i18n

fmontes:


src/api/services/login-service.ts
```diff
@@ -22,58 +19,59 @@ import {Subject} from 'rxjs/Subject';
...
-        this.loginAsUrl = `${apiRoot.baseUrl}api/v1/users/loginas/userid`;
+        this._loginAsUsersList$ = <Subject<User[]>>new Subject();
         this.loginAsUserList = [];
-        this.logoutAsUrl = `${apiRoot.baseUrl}api/v1/users/logoutas`;
-        this.logoutURL = `${apiRoot.baseUrl}api/v1/logout`;
-        this.recoverPasswordURL = `${apiRoot.baseUrl}api/v1/forgotpassword`;
-        this.serverInfoURL = `${apiRoot.baseUrl}api/v1/loginform`;
-        this.userAuthURL = `${apiRoot.baseUrl}api/v1/authentication`;
-        this.userListUrl = `${apiRoot.baseUrl}api/v1/users/loginAsData`;
+        this.urls = {
```
fmontes: We should do this all the time.

src/view/components/common/login-as/login-as.html
```diff
@@ -9,11 +9,11 @@
...
             </div>
             <div class="form-group" *ngIf="needPassword">
-                <md-input name="password" id="login-as-password" type="password" placeholder="Password" floatingPlaceholder="false"   [required]="needPassword" ngModel></md-input>
+                <md-input name="password" id="login-as-password" type="password" placeholder="Password" floatingPlaceholder="false" [required]="needPassword" ngModel></md-input>
             </div>
             <div class="form-group">
-                <button md-raised-button [disabled]="!loginAsForm.valid">Change User</button>
-                <button md-button color="accent" (click)="close($event)">Cancel</button>
+                <button md-raised-button [disabled]="!loginAsForm.valid" messagekey>change</button>
+                <button md-button color="accent" (click)="close($event)" messagekey>cfancel</button>
```
oswaldogallango: error in cancel has an extra f ==> cfancel

fmontes: Done.

src/api/services/login-service.ts
freddyucv: I think that we should call the requestView method like this better:

requestView({
   url: 'v1/authentication',
  ...
})

I did a change in CoreWebService that allow it

src/api/services/dotcms-events-service.ts
```diff
@@ -24,16 +24,12 @@ export class DotcmsEventsService {
...
         this.protocol = dotcmsConfig.getWebsocketProtocol();
         this.baseUrl = dotcmsConfig.getWebsocketBaseUrl();
         this.endPoint = dotcmsConfig.getSystemEventsEndpoint();
 
-        if (loginService.loginUser) {
-            this.connectWithSocket(this.loginService.loginUser);
-        }
-
-        this.loginService.loginUser$.subscribe(user => this.connectWithSocket(user));
+        loginService.watchUser(this.connectWithSocket.bind(this));
```
freddyucv: I think is better:

loginService.watchUser( () => this.connectWithSocket() );

src/api/services/login-service.ts
```diff
@@ -22,58 +19,59 @@ import {Subject} from 'rxjs/Subject';
...
 
-    get isLoginAs$(): Subject<boolean> {
-        return this._isLoginAs$.asObservable();
+    get loginAsUsersList$(): Subject<User[]> {
+        return this._loginAsUsersList$.asObservable();
     }
 
-    get loginAsUsers$(): Subject<User[]> {
-        return this._loginAsUsers$.asObservable();
+    get auth$(): Observable<Auth> {
```
freddyucv: I think that we shouldn't have an Observable for auth, we should have two Observable instead, one for the login user and one for the loginAS user, the first one should emits  in both case (login and loginAs), this way the LoginAs logic is encapsulated in the LoginService

src/view/components/common/login-as/login-as.ts
```diff
@@ -17,20 +19,29 @@ import {Router} from '@ngrx/router';
...
+        this.messageMapSubscription = this.messageService.getMessages(this.i18nKeys).subscribe(res => {
+            this.i18nMessages = _.pick(res, this.i18nKeys);
+        });
     }
 
     ngOnInit(): void {
-        this.loginService.loadLoginAsUsers();
-        this.loginService.loginAsUsers$.subscribe(data => {
+        this.loginService.loadLoginAsUsersList();
+        this.loginService.loginAsUsersList$.subscribe(data => {
```
freddyucv: Maybe we could have just one method, the loginAsUsersList$ could check if the list is already loaded and if not then load it.

src/view/components/common/login-as/login-as.ts
```diff
@@ -17,20 +19,29 @@ import {Router} from '@ngrx/router';
...
+    private i18nMessages = {};
+    private messageMapSubscription;
     private needPassword: boolean = false;
+    private userLists: Array<User>;
 
-    constructor(private loginService: LoginService, private router: Router,
-                @Inject('dotcmsConfig') private dotcmsConfig: DotcmsConfig) {
+    constructor(private loginService: LoginService, private router: Router, private messageService: MessageService) {
+        this.messageMapSubscription = this.messageService.getMessages(this.i18nKeys).subscribe(res => {
+            this.i18nMessages = _.pick(res, this.i18nKeys);
```
freddyucv: We could create a BaseComponent class and move this code to it, something like:

class BaseComponent{
   constructor(i18nKeys:string[], messageService: MessageService){
       this.messageService.getMessages(i18nKeys).subscribe(res => {
          ....
       });
   }
} 

and to use it:

class MyComponent extends BaseComponents{
     constructor(){
          super(['key-1', 'key-2']);
     }
}

src/view/components/common/login-as/login-as.ts
```diff
@@ -17,20 +19,29 @@ import {Router} from '@ngrx/router';
...
     ngOnInit(): void {
-        this.loginService.loadLoginAsUsers();
-        this.loginService.loginAsUsers$.subscribe(data => {
+        this.loginService.loadLoginAsUsersList();
+        this.loginService.loginAsUsersList$.subscribe(data => {
             this.userLists = data;
         });
     }
 
+    ngOnDestroy() {
```
freddyucv: Maybe we could move it to BaseComponent too

src/view/components/common/toolbar-user/toolbar-user.ts
```diff
@@ -18,29 +15,23 @@ import {DotcmsConfig} from '../../../../api/services/system/dotcms-config';
...
-        this.user = this.loginService.loginUser;
-        this.loginService.loginUser$.subscribe((user) => {
-            this.user = user;
-        });
-        this.isLoggedAs = this.dotcmsConfig.configParams.loginAsUser ? true : false;
-        this.loginService.isLoginAs$.subscribe(data => {
-            this.isLoggedAs = data;
+        this.auth = this.loginService.auth;
+        this.loginService.auth$.subscribe((auth) => {
+            this.auth = auth;
```
freddyucv: We should use the new watchUser method

src/view/components/my-account-component/dot-my-account-component.ts
```diff
@@ -27,38 +26,62 @@ export class MyAccountComponent {
...
+        if (this.loginService.auth.user) {
+            this.loadUser(this.loginService.auth.user);
+        }
 
-    save(): void {
+        this.loginService.auth$.subscribe(auth => {
+            if (auth.user) {
+                this.loadUser(auth.user);
+            }
+        });
```
freddyucv: We should use the new watchUser method

src/api/services/dotcms-events-service.ts
fmontes: Meh... I like it better this way, because we are explicitly passing the context where to execute that method. 

So 50-50... any other opinion people?

src/view/components/common/toolbar-user/toolbar-user.ts
fmontes: Yup... I miss this one.

src/api/services/login-service.ts
fmontes: Done: https://github.com/dotCMS/core-web/pull/186/commits/c61628a7be24e926a37712dde16b4586387cdc8c

src/view/components/common/login-as/login-as.ts
fmontes: Done: https://github.com/dotCMS/core-web/pull/186/commits/a92d671f644ed7f851a8f8102fe7fbf271caf80c

src/view/components/common/toolbar-user/toolbar-user.ts
fmontes: Done: https://github.com/dotCMS/core-web/pull/186/commits/cb7318198645d88452639a4f854bc572e8f5eeaa

src/api/services/bundle-service.ts
```diff
@@ -25,26 +25,24 @@ export interface IPublishEnvironment {
...
-    this._pushEnvironementsUrl = `${this._apiRoot.baseUrl}api/environment/loadenvironments/roleId`
-    this._pushRuleUrl = `${this._apiRoot.baseUrl}DotAjaxDirector/com.dotcms.publisher.ajax.RemotePublishAjaxAction/cmd/publish`
-
-
+  private _bundlesAry:IBundle[] = [];
+  private _environmentsAry:IDBEnvironment[] = [];
+
+  constructor(public _apiRoot:ApiRoot, _http:Http) {
+    super(_apiRoot, _http);
+    this._bundleStoreUrl = `${this._apiRoot.baseUrl}api/bundle/getunsendbundles/userid`;
```
jdotcms: why do we use underscore for the class properties?

src/api/services/dotcms-events-service.ts
jdotcms: I think both are ok, do not see much cons or pros. #upToYou

src/api/services/messages-service.ts
```diff
@@ -0,0 +1,85 @@
...
+    private _messageMap$: Subject<any> = new Subject<any>();
+    private doMessageLoad;
+    private i18nUrl: string;
+    private lang: string;
+    private messageKeys: Array;
+    private messagesLoaded: any;
+
+    constructor(apiRoot: ApiRoot, http: Http, loginService: LoginService) {
+        super(apiRoot, http);
+        this.i18nUrl = 'v1/languages/i18n';
```
jdotcms: the url could be encapsulated in an interface/factory, if the version changed we would like to avoid to change on all url over the app the v1 or even worst change in some of them, and some of them not.

Eventually we have to think in the possibility to support components by version

src/view/components/common/_base/base-component.ts
```diff
@@ -0,0 +1,24 @@
...
+import {Component, ViewEncapsulation} from '@angular/core';
+import {MessageService} from '../../../../api/services/messages-service';
+
+@Component({
+    encapsulation: ViewEncapsulation.Emulated,
+    moduleId: __moduleName,
+    selector: 'base'
+})
+
+export class BaseComponent {
```
jdotcms: Like it

src/view/components/common/login-as/login-as.ts
fmontes: Done

src/view/components/my-account-component/dot-my-account-component.ts
fmontes: Done

src/api/services/bundle-service.ts
fmontes: Old code will be a refactor.

src/api/services/messages-service.ts
fmontes: @jdotcms agreed, can you create a task and pass it to @jgambarios please?

src/api/services/dotcms-events-service.ts
fmontes: Sorry @freddyucv :)

src/api/services/login-service.ts
fmontes: I talked to @freddyucv and wit the ```watchUser``` change it's not worthy to do this.

PR #187, opened 2016-09-12
dotCMS/core#9642 logout if session expired

oswaldogallango:


src/view/components/common/iframe-legacy/Iframe-legacy-component.ts
```diff
@@ -65,4 +66,22 @@ export class IframeLegacyComponent extends SiteChangeListener {
...
+     * Validate if the iframe window is send to the login page after jsessionid expired
+     * then logout the user from angular session
+     */
+    checkSessionExpired(): void {
+        if (this.iframeElement && this.iframeElement.contentWindow) {
+            let currentPath = this.iframeElement.contentWindow.location.pathname;
+
+            if (currentPath.indexOf('/c/portal_public/login') != -1) {
+                this.loginService.logOutUser().subscribe(data => {
+                    this.router.go('/public/login');
```
freddyucv: I think that the logOutUser() go to '/public/login'

src/view/components/common/iframe-legacy/Iframe-legacy-component.ts
```diff
@@ -65,4 +66,22 @@ export class IframeLegacyComponent extends SiteChangeListener {
...
+
+    /**
+     * Validate if the iframe window is send to the login page after jsessionid expired
+     * then logout the user from angular session
+     */
+    checkSessionExpired(): void {
+        if (this.iframeElement && this.iframeElement.contentWindow) {
+            let currentPath = this.iframeElement.contentWindow.location.pathname;
+
+            if (currentPath.indexOf('/c/portal_public/login') != -1) {
```
freddyucv: This is weird, maybe here:

https://github.com/dotCMS/core/blob/master-3.7/src/com/liferay/portal/struts/PortalRequestProcessor.java#L340

We could send a Http error code if the request come from a angular UI, and then in the iFrameComponent listen the onError event, something like:

this.iframeElement.onerror = () => //Go to login

I don't know if that work but maybe we can try

oswaldogallango: Unfortunately, this is not easy to accomplish because the variable indicating that we are using an iframe is pass only one time as a request parameter, after that is saved in session, so if the user is navigating inside the iframe the parameter is kept in session but at this point, the session doesn't exist anymore

PR #190, opened 2016-09-14
Issue 9734 error changing user password from myaccount

oswaldogallango:


oswaldogallango:
this pull requires also https://github.com/dotCMS/core/pull/9791

PR #193, opened 2016-09-20
dotCMS/core#9300 add missing iframe name attribute

oswaldogallango:


src/view/components/common/iframe-legacy/iframe-legacy-component.html
```diff
@@ -1,6 +1,7 @@
 <md-spinner color="accent" *ngIf="loadingInProgress" class="iframe-spinner"></md-spinner>
 <iframe   width="100%"
           height="100%"
+          name="detailFrame"
```
fmontes: +1

PR #201, opened 2016-10-03
dotCMS/core#9798 listening the events SAVE_BASE_CONTENT_TYPE,UPDATE_B…

freddyucv:
…ASE_CONTENT_TYPE and DELETE_BASE_CONTENT_TYPE

### DIFF
--- a/src/api/services/contentlet-service.ts
+++ b/src/api/services/contentlet-service.ts
@@ -20,24 +20,21 @@ export class ContentletService extends CoreWebService {
         loginService.watchUser(this.loadContentTypes.bind(this));
 
         dotcmsEventsService.subscribeTo('SAVE_BASE_CONTENT_TYPE').pluck('data').subscribe( contentTypeView => {
-            console.log('SAVE_BASE_CONTENT_TYPE', contentTypeView);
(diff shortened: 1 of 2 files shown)


src/api/services/contentlet-service.ts
```diff
@@ -7,17 +7,42 @@ import {ApiRoot} from '../persistence/ApiRoot';
...
     private _structureTypeView$: Subject<StructureTypeView[]> = new Subject<StructureTypeView[]>();
 
-    constructor(apiRoot: ApiRoot, http: Http, loginService: LoginService) {
+    constructor(apiRoot: ApiRoot, http: Http, loginService: LoginService, dotcmsEventsService: DotcmsEventsService) {
         super(apiRoot, http);
 
         loginService.watchUser(this.loadContentTypes.bind(this));
-        loginService.logoutAs().subscribe( () => this.loadContentTypes());
+
+        dotcmsEventsService.subscribeTo('SAVE_BASE_CONTENT_TYPE').pluck('data').subscribe( contentTypeView => {
```
jdotcms: This is pretty cool and extensible (liked)
The only thing I can see is that we should introduce a logging library, the console.log should be not showed on production mode, just on developer mode.

Thanks

PR #211, opened 2016-10-10
dotCMS/core#9835 QA feedback round 1

fmontes:


gulpfile.js
```diff
@@ -104,6 +105,14 @@ var project = {
                 inlinejs: [ uglify() ],
                 inlinecss: [ minifyCss(), 'concat' ]
               }))
+              .pipe(replace({
```
fmontes: This is something we need to do because we run the app in different folders in dev and prod.

src/index.html
```diff
@@ -1,7 +1,7 @@
 <!doctype html>
 <html lang="en">
 <head>
-  <base href="/html/ng/">
```
fmontes: https://github.com/dotCMS/core-web/pull/211/files#r82599764

PR #213, opened 2016-10-10
Issue 9428 websocket reconnection

freddyucv:


src/api/services/dotcms-events-service.ts
```diff
@@ -16,6 +16,7 @@ export class DotcmsEventsService {
     private endPoint: String;
 
     private subjects: Subject<any>[] = [];
+    private TIME_TO_WAIT_TO_RECONNECT: number = 1000;
```
oswaldogallango: I think this value should be taken from some dotCMS property file and not hard coded.  Allowing the user to have the option to change the value

freddyucv: done

PR #220, opened 2016-11-01
dotCMS/core#9943

oswaldogallango:


jgambarios:
Closed as we are going to implemented with another and more generic approach.

PR #229, opened 2016-11-16
dotCMS/core#10081 Style side navigation to match the design

fmontes:


gulpfile.js
```diff
@@ -170,6 +170,7 @@ var project = {
           gulp.src('./src/**/*.scss')
               .pipe(sourcemaps.init())
               .pipe(sass({outputStyle: config.buildTarget === 'dev' ? 'expanded' : 'compressed'}))
+              .on('error', function() {})
```
fmontes: This is so the watch mode don't break when there is an sass error.

src/view/components/common/main-navigation/main-navigation.ts
```diff
@@ -12,13 +13,22 @@ import {RoutingService, Menu} from '../../../../api/services/routing-service';
 export class MainNavigation {
 
     private menuItems: Menu[];
+    private menuIcons = {
+        'Home': 'home',
```
fmontes: Map for icons in font-awesome, the key is the tab name from the menu service.

src/view/components/common/main-navigation/main-navigation.ts
```diff
@@ -12,13 +13,22 @@ import {RoutingService, Menu} from '../../../../api/services/routing-service';
...
+        'System': 'cog',
+    };
 
     constructor(routingService: RoutingService) {
         if (routingService.menus) {
             this.menuItems = routingService.menus;
         }
 
         routingService.menusChange$.subscribe(menu => {
+            console.log(menu);
```
fmontes: Ooops!

PR #231, opened 2016-11-21
Issue 10018 notify site browser changes

freddyucv:


src/api/services/routing-private-auth-service.ts
```diff
@@ -15,32 +15,47 @@ export class RoutingPrivateAuthService implements CanActivate {
     canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<boolean> {
 
         return new Promise(resolve => {
```
fmontes: Why is this a Promise and not an Observable?

src/api/services/routing-private-auth-service.ts
```diff
@@ -15,32 +15,47 @@ export class RoutingPrivateAuthService implements CanActivate {
...
-            if (this.loginService.isLogin) {
-                resolve(this.checkAccess(state.url));
-            } else {
+            if (!this.loginService.isLogin) {
```
fmontes: Also I don't get this... I mean ```if (this.loginService.isLogin || !this.loginService.isLogin)``` we still call ```this.checkAccessLoginUser```.

So it's like saying if user is login or check the access of the user. It doesn't make sense.

src/api/services/routing-private-auth-service.ts
```diff
@@ -15,32 +15,47 @@ export class RoutingPrivateAuthService implements CanActivate {
...
                 });
+            });
+        } else {
+            this.router.goToLogin();
+            resolve(false);
+        }
+    }
+
+    private checkAccess(url: string): Promise<boolean> {
+        return new Promise( resolve => {
```
fmontes: I don't understand why we have ```Promise``` all over the place.

PR #250, opened 2017-01-20
dotCMS/core#10059 reduced path for ng components

oswaldogallango:


src/api/services/routing-root-auth-service.ts
```diff
@@ -7,6 +7,9 @@ import {DotcmsConfig} from './system/dotcms-config';
 import {LoginService} from './login-service';
 import {DotRouterService} from './dot-router-service';
 
+/**
+ * TODO: This service is not used any more. Maybe should be deleted
```
jdotcms: add some comment here :)

oswaldogallango: the comment is already included it says: "TODO:This service is not used any more. Maybe should be deleted"

src/view/components/common/main-navigation/main-navigation.html
```diff
@@ -4,13 +4,13 @@
             <accordion-group [heading]="menu.tabName" [icon]="menuIcons[menu.tabName]">
                 <div class="main-nav__submenu">
                     <template ngFor let-menuItem [ngForOf]="menu.menuItems">
-                        <a [routerLink]="[menuItem.id]" *ngIf="menuItem.url && !menuItem.angular && !menuItem.ajax">
+                        <a [routerLink]="['/c/' + menuItem.id]" *ngIf="menuItem.url && !menuItem.angular && !menuItem.ajax">
```
freddyucv: Maybe is better concat this into typescript code instead.

Maybe into RoutingService, something like (when we iterate the menu items)
...
menuItem.routerLink = '/c/' + menuItem.id;
...

and then use it into the html:

<a [routerLink]="[menuItem.routerLink]" *ngIf="menuItem.url && !menuItem.angular && !menuItem.ajax">

PR #255, opened 2017-01-25
dotCMS/core#10382 fixed click navigation item to reload same page

oswaldogallango:


src/view/components/common/iframe-legacy/Iframe-legacy-component.ts
```diff
@@ -91,4 +102,17 @@ export class IframeLegacyComponent extends SiteChangeListener {
...
         }
     }
+
+    /**
+     * Force to reload the current iframe component portlet,
+     * with the specified portlet url
+     * @param url Url of the portlet to display
+     */
+    reloadIframePortlet(url: string): void {
+        if(url !== undefined && url !== '') {
```
freddyucv: it should work just with:

if(!url && url !== '') {

src/view/components/common/iframe-legacy/Iframe-legacy-component.ts
```diff
@@ -91,4 +102,17 @@ export class IframeLegacyComponent extends SiteChangeListener {
...
+            let urlSplit = url.split('/');
+            let id = urlSplit[urlSplit.length - 1];
+            this.iframe = this.loadURL(this.routingService.getPortletURL(id) + '&in_frame=true&frame=detailFrame');
+        }
```
freddyucv: Why don't we just do the following?

this.iframe = this.loadURL(url + '&in_frame=true&frame=detailFrame');

oswaldogallango: that is not going to work because that url is the one we use to get real portlet url. I mean the url are like /c/site-browser and not the real one with the jsp we use in the iframe. the real url we get it from the this.routingService.getPortletURL(id)

src/view/components/common/iframe-legacy/Iframe-legacy-component.ts
oswaldogallango: that change doesn't work. I tested and throw false when the url is set

PR #257, opened 2017-01-27
dotCMS/core#10457 Fix site selector width

fmontes:


.gitignore
```diff
@@ -12,6 +12,7 @@ staging/
 /typings/
 !staging/.gitkeep
 /semantic/
+src/**/*.css
```
fmontes: We don't want compile css in the repo, they are generated when we create an snapshot and we had to clean them up manually, so not they are ignored.

PR #272, opened 2017-02-15
Issue 10660 main navigation needs selected portlet highlight

freddyucv:


src/view/components/common/main-navigation/main-navigation.ts
```diff
@@ -22,13 +25,24 @@ export class MainNavigation {
...
+        });
     }
 
     /**
      * Change or refresh the portlets from the main menu
-     * @param link portlet url
+     * @param menuItem portlet url
      */
-    public gotToPage(link: string): void {
+    public gotToPage(menuItem: MenuItem): void {
```
oswaldogallango: why did you change the firm?. At the end you are only taking the menulink, I dont get why do you pass the whole menu item

PR #279, opened 2017-02-22
#10774 new loading screen

wezell:


wezell:
makes it this:

screen shot 2017-02-22 at 12 28 15 pm

PR #283, opened 2017-02-27
dotCMS/core#10770 On logging in, make default screen the first portlet in the first layout

fmontes:


src/api/services/routing-private-auth-service.ts
```diff
@@ -13,18 +12,23 @@ export class RoutingPrivateAuthService implements CanActivate {
...
-                            if (!checkAccess) {
-                                this.router.goToMain();
-                            }
-                            obs.next(checkAccess);
-                        });
+                    this.dotcmsConfig.getConfig().subscribe(configParams => {
+                        if (state.url.indexOf('home') > -1) {
+                            this.routingService.menusChange$.subscribe(res => {
+                                this.router.goToURL(`/c/${this.routingService.firstPortlet}`);
+                                obs.next(false);
```
freddyucv: maybe we can do something like:

 this.router.gotoFirstPortlet();

fmontes: Try to do it, but somehow when you add routingService to dotroutingService explode!

PR #370, opened 2017-06-16
Issue 11817 add date colum to listing datatable

woontsoudev:


src/app/view/components/listing-data-table/listing-data-table.component.html
```diff
@@ -11,7 +11,8 @@
              (onLazyLoad)="loadDataPaginationEvent($event)"
              [(selection)]="selectedItems"
              (onRowClick)="handleRowClick($event)"
-             [globalFilter]="gf">
+             [globalFilter]="gf"
+             sortField="{{sortField}}">
```
freddyucv: This input's name is 'defaultSortField', it should not work in this way

src/app/view/components/listing-data-table/listing-data-table.component.ts
```diff
@@ -17,27 +18,38 @@ export class ListingDataTableComponent extends BaseComponent {
...
     }
-
     ngOnInit(): void {
         this.dotcmsConfig.getConfig().subscribe(configParams => {
             this.paginatorRows = configParams.paginatorRows;
             this.paginatorLinks = configParams.paginatorLinks;
             this.loadData(this.paginatorRows, -1);
         });
+
+        this.sortField = this.defaultSortField;
```
freddyucv: Why we need both 'sortField' and defaultSortField?

src/app/view/components/listing-data-table/listing-data-table.component.ts
```diff
@@ -62,10 +74,12 @@ export class ListingDataTableComponent extends BaseComponent {
      * @memberof ListingDataTableComponent
      */
     loadData(limit: number, offset: number, sortField?: string, sortOrder?: number, query?: string): void {
-        this.crudService.loadData(this.url, limit, offset, sortField,
-        sortOrder < 0 ? OrderDirection.DESC : OrderDirection.ASC, query)
-            .subscribe( response => {
-                this.items = response.items;
+        if (sortField) {
+            sortField = this.defaultSortOrder;
```
freddyucv: This is not right, it should be:
if (!sortField) {
    sortField = this. defaultSortField;
}

or ...

sortField = !sortField ? this. defaultSortField : sortField;

src/app/view/components/listing-data-table/listing-data-table.component.ts
```diff
@@ -17,27 +18,38 @@ export class ListingDataTableComponent extends BaseComponent {
...
             this.paginatorLinks = configParams.paginatorLinks;
             this.loadData(this.paginatorRows, -1);
         });
+
+        this.sortField = this.defaultSortField;
+    }
+
+    ngOnChanges(changes): void {
+        if (changes.columns) {
+            this.dateColumns = changes.columns.currentValue.filter(column => column.format === 'date');
```
freddyucv: Here we can use a read only value, something like:

readonly DATE_FORMAT = 'date';

and ...
...
this.dateColumns = changes.columns.currentValue.filter(column => column.format === this.DATE_FORMAT);
...

Also, I dont know if handle this property as a String is a good idea, do we expect others values for this property in the future?, I don't think so, so I think we should use a boolean for it and call it something like 'dateFormat' (at least for now)

src/app/view/components/listing-data-table/listing-data-table.component.ts
```diff
@@ -17,27 +18,38 @@ export class ListingDataTableComponent extends BaseComponent {
     @Input() columns: DataTableColumn[];
     @Input() url: string;
     @Input() actionHeaderOptions: ActionHeaderOptions;
+    @Input() buttonActions: ButtonAction[] = [];
+    @Input() defaultSortOrder: string;
```
freddyucv: I think that this property is to set the default order direction (ASC or DESC), so it should be a 'OrderDirection' not a string

src/app/view/components/_common/pattern-library/pattern-library.html
```diff
@@ -50,9 +50,11 @@ <h3 class="section__title">Action button</h3>
...
             <h3 class="section__title">Listing table</h3>
             <!--<listing-data-table url="v1/contenttype"
-                                        [columns]="contentTypeColumns"
-                                        [buttonActions]="buttonActions">
-            </listing-data-table>-->
+                                [columns]="contentTypeColumns"
+                                [buttonActions]="buttonActions"
+                                [defaultSortOrder]="'mod_date'"
+                                [defaultSortField]="'mod_date'">
+            </listing-data-table>  -->
```
freddyucv: Why is this code commented?

src/app/view/components/listing-data-table/listing-data-table.component.ts
```diff
@@ -17,27 +18,41 @@ export class ListingDataTableComponent extends BaseComponent {
...
     }
 
     ngOnInit(): void {
         this.dotcmsConfig.getConfig().subscribe(configParams => {
             this.paginatorRows = configParams.paginatorRows;
             this.paginatorLinks = configParams.paginatorLinks;
-            this.loadData(this.paginatorRows, -1);
         });
+
+        this.defaultSortOrder = (this.sortOrder === 'desc' ? -1 : 1);
```
fmontes: You don't need this here... you can pass this directly in the template

```[sortOrder]="sortOrder === 'desc' ? -1 : 1"```

Because if you don't... and the ```sortOrder``` param changes you need to update the ```defaultSortOrder```.

File: src/app/view/components/listing-data-table/listing-data-table.component.ts [hunk 8]
```diff
@@ -17,27 +18,41 @@ export class ListingDataTableComponent extends BaseComponent {
...
+        this.dataFormat = 'date';
```
fmontes: I think you don't need this, because I think you are "double filtering"... read the next comments.

src/app/view/components/listing-data-table/listing-data-table.component.ts
```diff
@@ -17,27 +18,41 @@ export class ListingDataTableComponent extends BaseComponent {
...
+    }
+
+    ngOnChanges(changes): void {
+        if (changes.columns.currentValue) {
+            this.dateColumns = changes.columns.currentValue.filter(column => column.format === this.dataFormat);
```
fmontes: Here, you have an array of columns where all the ```format``` property is ```'date'```

src/app/view/components/listing-data-table/listing-data-table.component.ts
```diff
@@ -81,6 +98,17 @@ export class ListingDataTableComponent extends BaseComponent {
         return col.textAlign ? col.textAlign :
             (this.items && this.items[0] && typeof this.items[0][col.fieldName] === 'number') ? 'right' : 'left';
     }
+
+     private formatData(items: any[]): any[] {
+        return items.map((item) => {
+            this.dateColumns.forEach((col) => {
+                if (col.format === this.dataFormat) {
```
fmontes: And because ```this.dateColumns``` is only "date" columns, you don't need this if.

src/app/view/components/listing-data-table/listing-data-table.component.ts
```diff
@@ -62,10 +77,12 @@ export class ListingDataTableComponent extends BaseComponent {
      * @memberof ListingDataTableComponent
      */
     loadData(limit: number, offset: number, sortField?: string, sortOrder?: number, query?: string): void {
-        this.crudService.loadData(this.url, limit, offset, sortField,
-        sortOrder < 0 ? OrderDirection.DESC : OrderDirection.ASC, query)
-            .subscribe( response => {
-                this.items = response.items;
+        sortField = sortField || this.sortField;
+        sortOrder = sortOrder || (this.sortOrder === 'desc' ? -1 : 1);
```
fmontes: I didn't see this before but... here instead of this, you can pass the comparation that Freddy did in the calling of the method:

```OrderDirection.DESC : OrderDirection.ASC```

File: tslint.json [hunk 12]
```diff
@@ -11,6 +11,7 @@
             "check-space"
         ],
         "curly": true,
+        "dom.iterable": true,
```
fmontes: I already add this, is in master.

src/app/view/components/_common/pattern-library/pattern-library.html
fmontes: I did this, because I have a task to update the PL.

PR #373, opened 2017-06-22
dotCMS/core#11844 Add icons option to content types listing

fmontes:


src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -49,72 +57,101 @@ export class ContentTypesPortletComponent extends BaseComponent {
...
-                        command: ($event) => {
+                        command: $event => {
                             this.createContentType('file', $event);
                         },
                         icon: 'fa-file-o',
                         label: this.i18nMessages['File']
                     },
                     {
-                        command: ($event) => {
+                        command: $event => {
```
jdotcms: Not sure but I think curly bracket are not needed, such as $even => this.createContentType('page', $event)

fmontes: If you do it without the curly brackets it will ```return``` its the equivalent of doing:

```
$event => {
  return this.createContentType('page', $event);
}
```

And we don't what that.

src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -49,72 +57,101 @@ export class ContentTypesPortletComponent extends BaseComponent {
...
         this.contentTypeColumns = [
-            {fieldName: 'name', header: this.i18nMessages['Structure-Name'], width: '35%', sortable: true},
-            {fieldName: 'velocityVarName', header: this.i18nMessages['Variable'], width: '10%'},
-            {fieldName: 'description', header: this.i18nMessages['Description'], width: '35%'},
-            {fieldName: 'nEntries', header: this.i18nMessages['Entries'], width: '10%'},
-            {fieldName: 'mod_date', header: 'Last Edit Date', width: '10%', format: 'date', sortable: true}
+            {
+                fieldName: 'name',
+                header: this.i18nMessages['Structure-Name'],
+                icon: (item: any): string => this.contentTypesInfoService.getIcon(item.type),
```
freddyucv: I think  you can change 'any' by 'ContentType' here

fmontes: Can't do... item is way different than 'ContentType', the endpoint return different things when you do listing or get by id.

PR #374, opened 2017-06-26
Dot cms/core#10309 define metadata for pagination in rest json responses

freddyucv:


src/app/api/services/paginator/paginator.service.ts
```diff
@@ -0,0 +1,203 @@
...
+@Injectable()
+export class PaginatorService {
+
+    public static readonly LINK_HEADER_NAME = 'Link';
+    public static readonly PAGINATION_PER_PAGE_HEADER_NAME = 'X-Pagination-Per-Page';
+    public static readonly PAGINATION_CURRENT_PAGE_HEADER_NAME = 'X-Pagination-Current-Page';
+    public static readonly PAGINATION_MAX_LINK_PAGES_HEADER_NAME = 'X-Pagination-Link-Pages';
+    public static readonly PAGINATION_TOTAL_ENTRIES_HEADER_NAME = 'X-Pagination-Total-Entries';
+
+   public static readonly OrderDirection = {
```
jdotcms: I think TS has enum, for your consideration.

src/app/api/services/paginator/paginator.service.ts
```diff
@@ -0,0 +1,203 @@
...
+        ASC: 'ASC',
+        DESC: 'DESC'
+    };
+
+    public links: Links = {};
+
+    public paginationPerPage: number;
```
jdotcms: shouldn't be private and set init on the constructor?

src/app/api/services/paginator/paginator.service.ts
```diff
@@ -0,0 +1,203 @@
...
+    public currentPage: number;
+    public maxLinksPage: number;
+    public totalRecords: number;
+
+    private _url: string;
```
jdotcms: are we using _pattern? I think on typed languages it shouldn't be using!

src/app/api/services/paginator/paginator.service.ts
```diff
@@ -0,0 +1,203 @@
...
+     */
+    public get(url?: string): Observable<any[]> {
+        let params: URLSearchParams = new URLSearchParams();
+
+        if (this.sortField) {
+            params.set('orderby', this.sortField);
+        }
+
+        if (this.sortOrder) {
+            params.set('direction', this.sortOrder === 1 ?
```
jdotcms: I would create something like PaginatorService.OrderDirection.fromSortOrder (sortOrder)

src/app/api/services/paginator/paginator.service.ts
```diff
@@ -0,0 +1,203 @@
...
+        return this.get(this.links.first);
+    }
+
+    /**
+     * request the  pageParam page
+     * @param {number} [pageParam=1] Page to request
+     * @returns {Observable<any[]>}
+     * @memberof PaginatorServic
+     */
+    public getPage(pageParam = 1): Observable<any[]> {
```
jdotcms: you could type the pageParam

src/app/api/services/paginator/paginator.service.ts
```diff
@@ -0,0 +1,203 @@
...
+            this.links[rel] = url;
+        });
+    }
+}
+
+interface Links {
+    first?: string;
+    last?: string;
+    next?: string;
+    'x-page'?: string;
```
jdotcms: should be use this like x-page, probably just xPage would be more standard and easier, if there is not any special case to use it as a 'x-page'

src/app/view/components/listing-data-table/listing-data-table.components.spec.ts
```diff
@@ -21,6 +23,10 @@ describe('Listing Component', () => {
   let de: DebugElement;
   let el: HTMLElement;
 
+  let initPagiationService = () => {
```
jdotcms: shouldn't be name initPaginationService?

src/app/view/components/listing-data-table/listing-data-table.components.spec.ts
```diff
@@ -80,10 +86,17 @@ describe('Listing Component', () => {
         columns: new SimpleChange(null, comp.columns, true)
     });
 
+    let dataList = fixture.debugElement.query(By.css('p-dataTable'));
```
jdotcms: this styles could be good to encapsulate in a interface or class or constants. When you need to change any of them it would be easier than search/replace

src/app/view/components/site-selector/dot-site-selector.component.spec.ts
```diff
@@ -38,39 +36,39 @@ describe('Site Selector Component', () => {
         ],
         providers: [
           {provide: SiteService, useValue: siteServiceMock},
-          IframeOverlayService
+          IframeOverlayService, PaginatorService
         ]
     });
 
     fixture = DOTTestBed.createComponent(SiteSelectorComponent);
     comp = fixture.componentInstance;
```
jdotcms: These names are not descriptive, it should be componentInstance, debugElement and nativeElement

src/app/view/components/site-selector/dot-site-selector.component.ts
```diff
@@ -94,9 +91,14 @@ export class SiteSelectorComponent implements ControlValueAccessor {
...
     paginateSites(filter = '',  page = 1): void {
-        this.siteService.paginateSites(filter, false, page, this.paginationPerPage).subscribe( sites => {
-            this.sitesCurrentPage = sites;
+        this.paginationService.filter = filter;
+        this.paginationService.getPage(page).subscribe( items => {
+            this.sitesCurrentPage = items;
             this.selectCurrentSite();
+
+            if (!this.totalRecords) {
+                this.totalRecords = this.paginationService.totalRecords;
```
jdotcms: I think this could change to this.totalRecords = this.totalRecords | this.paginationService.totalRecords;

src/app/api/services/paginator/paginator.service.ts
freddyucv: this is a standard in ts, you can see a example here:

https://github.com/Microsoft/TypeScript-Handbook/blob/3436aca106c664ee77691642c14b295c148cfc75/pages/Classes.md

src/app/api/services/paginator/paginator.service.ts
fmontes: In Typescript when you assign a value it will assign the type by default... in fact if you add the type tslint will tell you it's redundant... at least that's what happen in a variable or property, not sure in a param like this... but I guess it's the same thing.

PR #381, opened 2017-06-29
Bug fixes and improvements for dotCMS/core#11844 and dotCMS/core#11968

fmontes:


src/app/api/services/content-types-info/content-types-info.service.ts
```diff
@@ -77,6 +77,13 @@ export class ContentTypesInfoService {
     }
 
     private getItem(type: string, prop: string): string {
+        // Some endpoints have different name/labels for the same content types
+        if (type === 'HTMLPAGE' || type === 'htmlpage') {
```
jdotcms: type.toLowerCase() could be easier to compare

src/app/view/components/listing-data-table/action-header/action-header.ts
```diff
@@ -15,7 +15,7 @@ export class ActionHeaderComponent extends BaseComponent {
     @Input() options: ActionHeaderOptions;
     public dynamicOverflow = 'visible';
 
-    constructor( messageService: MessageService, private confirmationService: ConfirmationService) {
+    constructor(messageService: MessageService, private confirmationService: ConfirmationService) {
```
jdotcms: why messageService is not private

src/app/api/services/content-types-info/content-types-info.service.ts
fmontes: true that... didn't notice at the time.

src/app/view/components/listing-data-table/action-header/action-header.ts
fmontes: Because we use it in the html... in fact we have some issues about that, I'm reading about how to handler props and services that are use in the templates.

PR #387, opened 2017-07-10
dotCMS/core#11627 Fix error with the new vanity URl content types

freddyucv:


src/app/api/services/content-types-info/content-types-info.service.ts
```diff
@@ -77,19 +77,36 @@ export class ContentTypesInfoService {
...
+
+        return result;
+    }
+
+    /**
+     * Some endpoints have different name/labels for the same content types
+     * @param type
+     */
+    private getTypeName(type: string): string {
         if (type === 'HTMLPAGE' || type === 'htmlpage') {
```
jdotcms: This might be in a Map such as 
type -> typeName

so you only have to do something like

return  typeNameMap[type];

will be log(N) instead of O(N) since won't be iterative testing each case.

PR #391, opened 2017-07-20
dotCMS/core#11876 Style new site-selector component

fmontes:


src/app/portlets/content-types/create/content-types-create.component.ts
```diff
@@ -19,24 +19,30 @@ import { ContentTypesFormComponent } from '../form';
  */
 @Component({
     selector: 'content-types-creaate',
```
woontsoudev: Misspelling on 'creaate'

PR #411, opened 2017-09-07
dotCMS/core#12372 Bring back the loading indicator missing after ng-cli

luisrojascr:


src/index.html
```diff
@@ -20,11 +20,109 @@
...
+        }
+    </style>
 </head>
 
 <body>
-    <app-root></app-root>
+	<app-root></app-root>
+	
+	<!-- Loading Indicator -->
+	<div class="loading-wrapper">
```
fmontes: You need to put the ```<div class="loading-wrapper">``` inside the ```<app-root>```

Like:

```html
<app-root>
    <div class="loading-wrapper">
        // WHATEVER
    </div>
</app-root>
```

If not, when the app loads, the html for the loading will be in the dom.

src/index.html
```diff
@@ -20,11 +20,109 @@
...
+		<div class="logo">
```
fmontes: Make sure everything it's 4 spaces indented.

PR #413, opened 2017-09-07
dotCMS/core#12194 site-selector needs to adjust to it content

luisrojascr:


src/app/view/components/_common/searchable-dropdown/component/searchable-dropdown.component.ts
```diff
@@ -46,7 +46,7 @@ export class SearchableDropdownComponent extends BaseComponent
     @Input() rows: number;
     @Input() totalRecords: number;
     @Input() placeholder = '';
-
+    @Input() width: string;
```
freddyucv: you missed test, you have to do at least one test with this attribute

PR #414, opened 2017-09-08
dotCMS/core#12269 We are unable to see the site list after try to use…

luisrojascr:
… the site filter


src/app/view/components/_common/site-selector/site-selector.component.ts
```diff
@@ -116,9 +116,8 @@ export class SiteSelectorComponent implements ControlValueAccessor {
      * @memberof SiteSelectorComponent
      */
     getSitesList(filter = '',  offset = 0): void {
-        if (filter.length) {
-            this.paginationService.filter = filter;
-        }
+        // Set filter if undefined
```
fmontes: ```// Set filter even if empty string```

PR #417, opened 2017-09-11
dotCMS/core#12370 Fix iframe component error after migration to ng-cli

fmontes:


src/app/view/components/iframe-legacy/Iframe-legacy-component.ts
```diff
@@ -105,7 +109,9 @@ export class IframeLegacyComponent extends SiteChangeListener {
      * @param $event
      */
     hideLoadingIndicator($event): void {
-        this.loadingInProgress = false;
+        setTimeout(() => {
```
fmontes: This solution came from this article: https://blog.angularindepth.com/everything-you-need-to-know-about-the-expressionchangedafterithasbeencheckederror-error-e3fd9ce7dbb4

freddyucv: but, why were we getting the error in first place?, I mean the hideLoadingIndicator method is called when the iframe is complete loaded, this happened after the back response with the current frame so the I suppose that angular's life cycle is ended

fmontes: Angular lifecycle start with the child components and up to the parent, so looks like when the iframe load the cycle hasn't finished in the parent... or something like that.

PR #421, opened 2017-09-15
dotCMS/core#12080 Add the new Vanity URL Content Types for the Add Co…

luisrojascr:
…ntent Types button


src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.ts
```diff
@@ -45,11 +45,16 @@ export class ToolbarAddContenletBodyComponent {
...
-    private types: StructureTypeView[];
+    types: StructureTypeView[];
+    mainTypes: StructureTypeView[];
+    moreTypes: StructureTypeView[];
+    moreItems = [];
+    private MAIN_CONTENT_TYPES = ['CONTENT', 'WIDGET', 'FORM', 'FILEASSET', 'HTMLPAGE'];
     private recent: StructureTypeView[];
-    private structureTypeViewSelected: StructureTypeView[];
-    private showMore = false;
+    structureTypeViewSelected: StructureTypeView[];
```
jdotcms: why are you removing the private?

src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.ts
```diff
@@ -65,24 +70,47 @@ export class ToolbarAddContenletComponent extends BaseComponent {
...
-                        this.recent.push(structure);
-                    } else {
-                        structure.types.forEach(type => {
-                            this.routingService.addPortletURL(type.name, type.action);
-                        });
+
+            this.recent = structures.filter(this.isRecentContentType);
+
+            this.types = structures.filter(structure => {
+                return !this.isRecentContentType(structure);
```
jdotcms: isn't the return redundant?

src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.ts
```diff
@@ -65,24 +70,47 @@ export class ToolbarAddContenletComponent extends BaseComponent {
...
-            );
+                });
+            });
 
             this.nextRecent();
         });
     }
 
+    private isRecentContentType(type: StructureTypeView): boolean {
+        return type.name.startsWith('RECENT');
```
jdotcms: shouldn't be a constant?

src/app/view/components/_common/action-button/action-button.component.html
```diff
@@ -1,7 +1,9 @@
...
-    icon="fa-plus" 
+    icon="fa-plus"
     class="action-button" 
     (onClick)="command()" 
     [model]="model" 
-    [class.multiple-options]="model && model.length">
-</p-splitButton>
\ No newline at end of file
+    [class.multiple-options]="model && model.length"
+    [class.toolbar-add-contentlet__head-button]="flat">
```
fmontes: This class name should be descriptive like "flat" related.

src/app/view/components/_common/action-button/action-button.component.html
```diff
@@ -1,7 +1,9 @@
...
+</p-splitButton>
+<span *ngIf="label" class="showMore">{{ label }}</span>
```
fmontes: This class name it's not BEM

src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.html
```diff
@@ -1,15 +1,17 @@
 <dot-dropdown-component
+    #dropdown
```
fmontes: I think we don't need this, it was just for the test, right?

src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.scss
```diff
@@ -20,3 +20,77 @@
...
 }
 
+
+.toolbar-add-contentlet__head-button{
+    &.action-button{
+        background: none;
+        border: none;
+        box-shadow: none;
+        border-radius: 50%;
+        height: 32px;
```
fmontes: There is a var name for this.

src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.scss
```diff
@@ -20,3 +20,77 @@
...
+    
+        &:hover {
+            border: none;
+            box-shadow: none;
+            background: rgba(22, 22, 22, 0.1);
```
fmontes: There is a var for this.

src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.ts
fmontes: Private are not accessible from the template (yup weird) https://stackoverflow.com/questions/34574167/angular2-should-private-variables-be-accessible-in-the-template

src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.ts
fmontes: Arrow function only return when you do like:

```structures.filter(structure => !this.isRecentContentType(structure))```

But turns out that Typescript won't allow to do ```!``` like that.

File: src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.ts (hunk 3 above)
Comment (fmontes) [reply src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.ts]: Maybe :)

File: src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.ts [hunk 9]
```diff
@@ -65,24 +70,47 @@ export class ToolbarAddContenletComponent extends BaseComponent {
...
+                return !this.isRecentContentType(structure);
+            });
+
+            this.types.forEach(structure => {
+                structure.types.forEach(type => {
+                    this.routingService.addPortletURL(type.name, type.action);
+                });
+            });
+            this.mainTypes = this.getMainContentType(this.types);
+            this.moreTypes = this.getMoreContentTypes(this.types);
```
fmontes: We can use ```.map()``` to do this, let me know and I'll explain to you.

src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.html
luisrojascr: Right, I'm removing that.

src/app/view/components/_common/action-button/action-button.component.html
freddyucv: Also I think showMore is not a good name, maybe something like showLabel is better.

src/app/view/components/toolbar-add-contentlet/toolbar-add-contentlet.scss
fmontes: No there is not :)

PR #425, opened 2017-09-18
dotCMS/core#12614 Wrong credentials crashing the app

luisrojascr:


src/app/view/components/login/login-component/login-container.ts
```diff
@@ -60,10 +60,10 @@ export class LoginContainer {
                 },
                 error => {
                     if (
-                        error.response.status === HttpCode.BAD_REQUEST ||
-                        error.response.status === HttpCode.UNAUTHORIZED
+                        error.status === HttpCode.BAD_REQUEST ||
+                        error.status === HttpCode.UNAUTHORIZED
                     ) {
-                        this.message = error.errorsMessages;
+                        this.message = JSON.parse(error._body).errors[0].message;
```
jdotcms: shouldn't make sure that the errors is not empty before gets the index 0

PR #437, opened 2017-10-04
dotCMS/core#12665 Unable to add/view relationships from the new Conte…

freddyucv:
…nt Types portlet

### DIFF
--- a/src/app/api/services/dot-menu.service.ts
+++ b/src/app/api/services/dot-menu.service.ts
@@ -8,7 +8,7 @@ import { DotMenu, DotMenuItem } from '../../shared/models/navigation';
 export class DotMenuService {
     menu$: Observable<DotMenu[]>;
 
-    private urlMenus = 'v1/CORE_WEB/menu';
+    private urlMenus = 'v1/menu';
 
     constructor(private coreWebService: CoreWebService) {}
 
@@ -71,6 +71,17 @@ export class DotMenuService {
     reloadMenu(): Observable<DotMenu[]> {
         this.menu$ = null;
         return this.loadMenu();
+    }
+
+    /**
+     * Return the DotMenu's ID of a Portlet
+     * @param portletId MenuItems id
+     */
+    getDotMenuId(portletId: string): Observable<string> {
+        return this.loadMenu()
+            .flatMap((menus: DotMenu[]) => menus)
+            .find((menu: DotMenu) => menu.menuItems.some(menuItem => menuItem.id === portletId))
+            .map((menu: DotMenu) => menu.id);
     }
 
     private getMenuItems(): Observable<DotMenuItem> {
--- a/src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
+++ b/src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
@@ -31,6 +31,14 @@ class TestDotIframe {
     @Input() src: string;
 }
 
+@Component({
+    selector: 'dot-content-types-relationship-listing',
+    template: ''
+})
+class TestContentTypesRelationshipListingComponent {
+
+}
+
 describe('ContentTypesLayoutComponent', () => {
     let comp: ContentTypesLayoutComponent;
     let fixture: ComponentFixture<ContentTypesLayoutComponent>;
@@ -41,10 +49,11 @@ describe('ContentTypesLayoutComponent', () => {
 
         const messageServiceMock = new MockMessageService({
             'contenttypes.sidebar.components.title': 'Field Title',
-            'contenttypes.tab.header.fields': 'Fields Header Tab',
+            'contenttypes.tab.fields.header': 'Fields Header Tab',
             'contenttypes.sidebar.layouts.title': 'Layout Title',
-            'contenttypes.tab.header.permissions': 'Permissions Tab',
-            'contenttypes.tab.header.publisher.push.history': 'Push History',
+            'contenttypes.tab.permissions.header': 'Permissions Tab',
+            'contenttypes.tab.publisher.push.history.header': 'Push History',
+            'contenttypes.tab.relationship.header': 'Relationship'
         });
 
         DOTTestBed.configureTestingModule({
@@ -52,7 +61,8 @@ describe('ContentTypesLayoutComponent', () => {
                 ContentTypesLayoutComponent,
                 TestContentTypeFieldsList,
                 TestContentTypeFieldsRowList,
-                TestDotIframe
+                TestDotIframe,
+                TestContentTypesRelationshipListingComponent
             ],
             imports: [
                 TabViewModule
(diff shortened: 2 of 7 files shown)


src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
```diff
@@ -211,10 +215,55 @@ describe('ContentTypesLayoutComponent', () => {
...
+
+            this.relationshipTab = this.contentType.queryAll(By.css('p-tabPanel'))[1];
+        });
+
+        it('should has a Relationship tab', () => {
+            expect(this.relationshipTab).toBeDefined();
+        });
+
+        it('should has a right header', () => {
+            expect('Relationship').toBe(this.relationshipTab.componentInstance.header);
```
fmontes: ```expect(this.relationshipTab.componentInstance.header).toBe('Relationship');```

src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
```diff
@@ -211,10 +215,55 @@ describe('ContentTypesLayoutComponent', () => {
...
     });
+
+    describe('Relationship tab', () => {
+        beforeEach(() => {
+            this.contentType = fixture.debugElement.query(By.css('.content-type'));
+            comp.contentTypeId = '1';
+
+            fixture.detectChanges();
+
+            this.relationshipTab = this.contentType.queryAll(By.css('p-tabPanel'))[1];
```
fmontes: This is not a good way to select the "Relationship tab" because if we change it of position this will break, you should use an ID or a BEM class.

src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
```diff
@@ -211,10 +215,55 @@ describe('ContentTypesLayoutComponent', () => {
...
+
+        it('should has a right header', () => {
+            expect('Relationship').toBe(this.relationshipTab.componentInstance.header);
+        });
+
+        it('should has a iframe', () => {
+            comp.contentTypeId = '1';
+
+            fixture.detectChanges();
+            const pTabPanel = this.contentType.queryAll(By.css('p-tabPanel'))[1];
```
fmontes: Again... don't use array position as a selector.

src/app/portlets/content-types/layout/content-types-layout.component.ts
```diff
@@ -15,21 +15,25 @@ export class ContentTypesLayoutComponent extends BaseComponent implements OnChan
...
+            'contenttypes.tab.relationship.header',
         ], messageService);
     }
 
     ngOnChanges(changes): void {
         if (changes.contentTypeId.currentValue) {
             this.permissionURL = `/html/content_types/permissions.jsp?contentTypeId=${changes.contentTypeId.currentValue}&popup=true`;
             this.pushHistoryURL = `/html/content_types/push_history.jsp?contentTypeId=${changes.contentTypeId.currentValue}&popup=true`;
+            // tslint:disable-next-line:max-line-length
+            this.relationshipURL = `c/portal/layout?p_l_id=56fedb43-dbbf-4ce2-8b77-41fb73bad015&p_p_id=content-types&_content_types_struts_action=%2Fext%2Fstructure%2Fview_relationships&_content_types_structure_id=${changes.contentTypeId.currentValue}`;
```
fmontes: ```id=56fedb43-dbbf-4ce2-8b77-41fb73bad015``` <-- you sure that we can hardcode this in the url?

src/app/api/services/dot-menu.service.ts
```diff
@@ -73,6 +73,17 @@ export class DotMenuService {
...
 
+    /**
+     * Return the DotMenu's ID of a Portlet
+     * @param portletId MenuItems id
+     */
+    getDotMenuId(portletId: string): Observable<string> {
+        return this.loadMenu()
+            .flatMap((menus: DotMenu[]) => menus)
+            .find((menu: DotMenu) => menu.menuItems.some(menuItem => menuItem.id === portletId))
+            .map((menu: DotMenu) => menu.id);
```
fmontes: I think you can do a second ```.flatMap((menu: DotMenu) => menu.menuItems)``` (or a map) and that way you don't need to use a ```.some```

src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
```diff
@@ -31,40 +33,70 @@ class TestDotIframe {
     @Input() src: string;
 }
 
+@Component({
+    selector: '<test></test>',
```
fmontes: Should this be just 'test'?

src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
```diff
@@ -75,146 +107,131 @@ describe('ContentTypesLayoutComponent', () => {
...
+            fixture.detectChanges();
+
+            this.pTabPanel = fixture.debugElement.query(By.css('p-tabView .content-type__properties'));
         });
 
         it('should has a field panel', () => {
-            fixture.detectChanges();
             expect(this.pTabPanel).not.toBeNull();
-            expect('Fields Header Tab').toBe(this.pTabPanel.componentInstance.header);
+            expect(this.pTabPanel.componentInstance.header).toBe('Fields Header Tab');
```
fmontes: Just FYI: https://stackoverflow.com/questions/22413009/jasmine-javascript-testing-tobe-vs-toequal

src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
```diff
@@ -75,146 +107,131 @@ describe('ContentTypesLayoutComponent', () => {
...
             const fieldTitle = this.pTabPanel.query(By.css('.content-type__fields-sidebar-title'));
             const contentTypesFieldsList = this.pTabPanel.query(By.css('content-types-fields-list'));
 
-            expect('Field Title').toBe(fieldTitle.nativeElement.textContent);
+            expect(fieldTitle.nativeElement.textContent).toBe('Field Title');
             expect(contentTypesFieldsList).not.toBeNull();
         });
 
         it('should has a field row list', () => {
-            fixture.detectChanges();
```
fmontes: Nice x10

src/app/view/components/dot-navigation/dot-navigation.component.html
```diff
@@ -6,7 +6,7 @@
                     <ng-template ngFor let-menuItem [ngForOf]="menu.menuItems">
                         <a ripple (click)="onClick($event, menuItem.id)" [routerLink]="menuItem.menuLink" 
                             [class.main-nav__submenu--active]="isActive(menuItem.id)">
-                            {{menuItem.name}}
+                            {{menuItem.label}}
```
fmontes: Why?

src/app/api/services/dot-menu.service.ts
freddyucv: If I did it on this way I would get MenuItem for the last map like this:

.map((menuItem: MenuItem) => ...);

... and I need return the DotMenu's id.

src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
freddyucv: really no matter what selector you use, but maybe be should define a standard

src/app/portlets/content-types/layout/content-types-layout.component.spec.ts
freddyucv: Yes in this case really is the same

src/app/view/components/dot-navigation/dot-navigation.component.html
freddyucv: because it is really a label, it is different according the user's language

PR #447, opened 2017-10-14
Issue 12747 update content type form

fmontes:


src/app/portlets/content-types/edit/content-types-edit.component.html
```diff
@@ -1,17 +1,15 @@
-<content-type-layout [contentTypeId]="data ? data.id : null">
-    <content-types-form
+<dot-content-type-layout [contentTypeId]="data?.id">
```
fmontes: We don't need to do ```object && object.prop``` anymore :)

src/app/portlets/content-types/edit/content-types-edit.component.html
```diff
@@ -1,17 +1,15 @@
...
+    <dot-content-types-form
         #form
+        [fields]="fields"
         [data]="data"
-        [type]="contentTypeType"
```
fmontes: Less inputs are always best.

src/app/portlets/content-types/content-types-resolver.service.ts
```diff
@@ -0,0 +1,43 @@
+import { Injectable } from '@angular/core';
```
freddyucv: I think is better named it something like: content-types.resolver.ts, it is not really a service

src/app/portlets/content-types/content-types-resolver.service.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Resolve, ActivatedRouteSnapshot, RouterStateSnapshot } from '@angular/router';
+import { Observable } from 'rxjs/Observable';
+import { CrudService } from '../../api/services/crud';
+import { ContentType, CONTENT_TYPE_INITIAL_DATA } from './shared/content-type.model';
+import { ContentTypesInfoService } from '../../api/services/content-types-info';
+import { LoginService } from 'dotcms-js/dotcms-js';
+import { DotRouterService } from '../../api/services/dot-router-service';
+
+@Injectable()
+export class ContentTypeResolver implements Resolve<ContentType> {
```
freddyucv: doc

src/app/portlets/content-types/content-types-resolver.service.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+        spyOn(crudService, 'getDataById').and.returnValue(Observable.of({
+            fake: 'content-type',
+            object: 'right?'
+        }));
+        contentTypeResolver
+            .resolve(activatedRouteSnapshotMock)
+            .subscribe((fakeContentType: any) => {
+                expect(fakeContentType).toEqual({
+                    fake: 'content-type',
+                    object: 'right?'
```
freddyucv: are you sure that this expect always is executed?, I think you need a tick or a whenStable o something like that

src/app/portlets/content-types/content-types-resolver.service.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    it('should get and return null and go to home', () => {
+        activatedRouteSnapshotMock.paramMap.get = () => '123';
+        spyOn(dotRouterService, 'gotoPortlet');
+        spyOn(crudService, 'getDataById').and.returnValue(Observable.of(false));
+        contentTypeResolver
+            .resolve(activatedRouteSnapshotMock)
+            .subscribe((res: any) => {
+                expect(res).toBeNull();
```
freddyucv: are you sure that this expect always is executed?, I think you need a tick or a whenStable o something like that

src/app/portlets/content-types/content-types-resolver.service.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    it('should return a content type placeholder base on type', () => {
+        activatedRouteSnapshotMock.paramMap.get = param => {
+            return param === 'type' ? 'content' : false;
+        };
+        spyOn(dotRouterService, 'gotoPortlet');
+        spyOn(crudService, 'getDataById').and.returnValue(Observable.of(false));
+        contentTypeResolver
+            .resolve(activatedRouteSnapshotMock)
+            .subscribe((res: any) => {
+                expect(res).toEqual({
```
freddyucv: are you sure that this expect always is executed?, I think you need a tick or a whenStable o something like that

src/app/portlets/content-types/content-types-resolver.service.ts
```diff
@@ -0,0 +1,43 @@
...
+    constructor(
+        private contentTypesInfoService: ContentTypesInfoService,
+        private crudService: CrudService,
+        private loginService: LoginService,
+        private dotRouterService: DotRouterService
+    ) {}
+
+    resolve(route: ActivatedRouteSnapshot): Observable<ContentType> {
+        if (route.paramMap.get('id')) {
+            return this.crudService
```
freddyucv: maybe we might change this for a private method something like getContentType

src/app/portlets/content-types/content-types-resolver.service.ts
```diff
@@ -0,0 +1,43 @@
...
+                    if (contentType) {
+                        return contentType;
+                    } else {
+                        this.dotRouterService.gotoPortlet('/content-types-angular');
+                        return null;
+                    }
+                });
+        } else {
+            const type = route.paramMap.get('type').toUpperCase();
+            return Observable.of(
```
freddyucv: maybe we might change this for a private method something like return defaultContentType

src/app/portlets/content-types/edit/content-types-edit.component.html
freddyucv: cool

src/app/portlets/content-types/edit/content-types-edit.component.ts
```diff
@@ -22,120 +17,102 @@ import { DotConfirmationService } from '../../../api/services/dot-confirmation';
...
+                this.fields = this.data.fields;
+                this.form.resetForm();
+                this.location.replaceState(`/content-types-angular/edit/${this.data.id}`);
             });
     }
 
-    private handleFormSubmissionResponse(res: any): void {
-        this.form.resetForm();
+    private getCreateContentTypeData(value: any): ContentType {
+        return Object.assign({}, CONTENT_TYPE_INITIAL_DATA, value);
```
freddyucv: is this really necessary? isn't this do by the resolver?

src/app/portlets/content-types/content-types-resolver.service.spec.ts
fmontes: So here is the deal, because the ```resolve``` is async also the method he use internally... but we are mocking all the responses and they are coming inmediatelly, we could use ```fakeAsync``` but it's not necessary because all the responses are inmediatelly available when the subscribe it's called.

src/app/portlets/content-types/content-types-resolver.service.ts
fmontes: I'm following the same pattern as with the guards as it is in the [angular documentation](https://angular.io/guide/router) they do like: ```src/app/crisis-center/crisis-detail-resolver.service.ts```

src/app/portlets/content-types/content-types-resolver.service.spec.ts
fmontes: https://github.com/dotCMS/core-web/pull/447#discussion_r144920884

src/app/portlets/content-types/content-types-resolver.service.spec.ts
fmontes: https://github.com/dotCMS/core-web/pull/447#discussion_r144920884

src/app/portlets/content-types/content-types-resolver.service.ts
fmontes: Done

src/app/portlets/content-types/content-types-resolver.service.ts
fmontes: Yup, great this way we can have a clear code.

src/app/portlets/content-types/content-types-resolver.service.ts
fmontes: Yup, great this way we can have a clear code.

src/app/portlets/content-types/edit/content-types-edit.component.ts
fmontes: I change this, now the form control handle ALL the Content Type object even when some properties are not represent for actual fields in the HTML

hmoreras:
Feedback after QA: 

- When expanded edit button should be disabled - is not happening.
- Openthe form change something, Update, then open the form agin and the “Update” btn keep enable. Should be disable since nothing change since the last Update.
- while doing tab to change between fields, should be nice to ignore the disabled fields. 
- Cancel Label not loading.. can be my environment. 
- When fields are empty is hard to identify witch ones are enable/disable.

fmontes:
- When expanded edit button should be disabled - is not happening.
Done
- Open the form change something, Update, then open the form agin and the “Update” btn keep enable. Should be disable since nothing change since the last Update.
Done

- while doing tab to change between fields, should be nice to ignore the disabled fields.
I'll create an issue for this

- Cancel Label not loading.. can be my environment.
Done

- When fields are empty is hard to identify witch ones are enable/disable.
Need to work on this, maybe change the design to be a dotted line

PR #448, opened 2017-10-16
dotcms/core#12748 [NC] Create add rows component

luisrojascr:


src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,162 @@
...
+import { MessageService } from './../../../../api/services/messages-service';
+
+/**
+ * Display select columns row
+ *
+ * @export
+ * @class ContentTypeFieldsAddRowComponent
+ */
+@Component({
+    selector: 'content-type-fields-add-row',
```
fmontes: Prefix the selector with ```dot``` and rename it like ```dot-add-rows``` because we'll be using this in edit pages and not only content types... in fact we probably move it some place else in the near future.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,162 @@
...
+})
+export class ContentTypeFieldsAddRowComponent extends BaseComponent implements OnInit, OnDestroy {
+    rows: number[] = [1, 2, 3, 4];
+    row: number;
+    fieldRows: FieldRow[];
+    addRowSelected = 'collapsed';
+    selectedCol = 0;
+    disabledButton = false;
+    comboHotKeys: Hotkey;
+    @Output() selectedColums: EventEmitter<number> = new EventEmitter<number>();
```
fmontes: Called ```selectRows``` because "selected" translate means that is selected and this is an so should be a verb.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    ngOnDestroy(): void {
+        this.hotkeysService.remove(this.comboHotKeys);
+    }
+
+    /**
+     * Set columns active when mouse enter
+     * @param col
+     */
+    public mouseEnter(col: number) {
```
fmontes: By default methods are public so you don't need to add ```public``` to it.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,162 @@
...
+        this.comboHotKeys = new Hotkey(['left', 'right', 'enter', 'esc'], (event: KeyboardEvent, combo: string): boolean => {
+            switch (combo) {
+                case 'left':
+                    this.leftKeyboardSelect();
+                    break;
+                case 'right':
+                    this.rightKeyboardSelect();
+                    break;
+                case 'enter':
+                    this.row = this.setRowListChild(this.selectedCol).children.length;
```
fmontes: I don't understand what's going on with this, shouldn't you emit the ```selectedCol```?

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.html
```diff
@@ -0,0 +1,21 @@
+<div class="content-type-fields_add-row__container">
+    <div class="content-type-fields-add-row" 
+        [ngClass]="{'content-type-fields-add-row__expanded': addRowSelected === 'expanded', 'content-type-fields-add-row__collapsed': addRowSelected === 'collapsed'}">    
+        <button type="button" pButton flat 
+            label="{{i18nMessages['contenttypes.content.add_rows']}}" (click)="addRow()"
```
freddyucv: I think a better key is contenttypes.dropzone.rows.add

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+        el = de.nativeElement;
+
+        comp.ngOnInit();
+
+        fixture.detectChanges();
+
+        const addButton = de.nativeElement.querySelector('.content-type-fields-add-row button');
+        spyOn(comp, 'addRow');
+        addButton.click();
+        comp.addRowSelected = 'expanded';
```
freddyucv: why do you need set this here?

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+        testHotKeysMock.callback('esc');
+
+        expect(comp.addRowSelected).toBe('collapsed');
+    });
+
+    it('should be able to use left keyboard to highlight columns', () => {
+        testHotKeysMock.callback('left');
+
+        expect(comp.selectedCol).toBe(3);
+    });
```
freddyucv: you need another test with 'left keyboard' where the selectCol is != of 0

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    it('should be able to use right keyboard to highlight columns', () => {
+        testHotKeysMock.callback('right');
+
+        expect(comp.selectedCol).toBe(1);
+    });
```
freddyucv: You need another test with 'right keyboard' where the selectCol is equals to 3

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    it('should select columns number after click on li', () => {
+        spyOn(comp.selectedColums, 'emit');
+        comp.selectRow(3);
+
+        expect(comp.selectedColums.emit).toHaveBeenCalledWith(3);
+    });
```
freddyucv: Really you aren't testing the click, you are testing the selectRow, you have to make the click on the button for this test

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    it('should select columns number after use enter keyboard on li', () => {
+        spyOn(comp.selectedColums, 'emit');
+
+        testHotKeysMock.callback('enter');
+        comp.selectRow(comp.row);
```
freddyucv: why are you calling the 'selectRow' method here?

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,162 @@
...
+    templateUrl: './content-type-fields-add-row.component.html',
+})
+export class ContentTypeFieldsAddRowComponent extends BaseComponent implements OnInit, OnDestroy {
+    rows: number[] = [1, 2, 3, 4];
+    row: number;
+    fieldRows: FieldRow[];
+    addRowSelected = 'collapsed';
+    selectedCol = 0;
+    disabledButton = false;
+    comboHotKeys: Hotkey;
```
freddyucv: I think you don't need this attribute, because yo can get the HotKey from the HotkeysService

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
freddyucv: Really you select columns number so maybe selectColumns is a better name

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,162 @@
...
+            transition('expanded <=> collapsed', animate('300ms ease-in-out'))
+        ])
+    ],
+    styleUrls: ['./content-type-fields-add-row.component.scss'],
+    templateUrl: './content-type-fields-add-row.component.html',
+})
+export class ContentTypeFieldsAddRowComponent extends BaseComponent implements OnInit, OnDestroy {
+    rows: number[] = [1, 2, 3, 4];
+    row: number;
+    fieldRows: FieldRow[];
```
freddyucv: are you using this attribute?

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,162 @@
...
+            ),
+            transition('expanded <=> collapsed', animate('300ms ease-in-out'))
+        ])
+    ],
+    styleUrls: ['./content-type-fields-add-row.component.scss'],
+    templateUrl: './content-type-fields-add-row.component.html',
+})
+export class ContentTypeFieldsAddRowComponent extends BaseComponent implements OnInit, OnDestroy {
+    rows: number[] = [1, 2, 3, 4];
+    row: number;
```
freddyucv: I don't understand why do you need this attribute?

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,162 @@
...
+     * @memberof ContentTypeFieldsAddRowComponent
+     */
+    public addRow(): void {
+        this.addRowSelected = 'expanded';
+    }
+
+    private leftKeyboardSelect(): any {
+        this.selectedCol = this.selectedCol - 1;
+        if (this.selectedCol < 0) {
+            this.selectedCol = 3;
```
freddyucv: use a const attribute for this, something like:

MAX_COLUMNS = 3

src/app/test/hotkeys-service.mock.ts
```diff
@@ -0,0 +1,27 @@
+import { Observable } from 'rxjs/Observable';
+import { HotkeysService, Hotkey } from 'angular2-hotkeys';
+
+export class TestHotkeysMock {
```
freddyucv: doc

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.html
jdotcms: agree

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,162 @@
...
+    private setRowListChild(col: number): any {
+        return this.rowList.nativeElement.children[col];
+    }
+
+    private resetRow(): any {
+        this.selectedCol = 0;
+        this.addRowSelected = 'collapsed';
+    }
+
+    private setTooltipValue(col: number): string {
```
jdotcms: this kind of things could be handle easier to achieve with just an array associated to the column

const toolTipValueMap = new Array( 'contenttypes.content.one_column', ''contenttypes.content.two_columns'' ...);

the method will be return this.i18nMessages[this.toolTipValueMap[col]];

src/app/portlets/content-types/content-types.module.ts
```diff
@@ -112,9 +114,11 @@ import { DotTextareaContentModule } from '../../view/components/_common/dot-text
         RadioButtonModule,
         ReactiveFormsModule,
         SiteSelectorModule,
+        ContentTypeFieldsAddRowModule,
         SplitButtonModule,
         TabViewModule,
-        DotTextareaContentModule
+        DotTextareaContentModule,
+        TooltipModule
```
fmontes: You don't need this anymore.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    it('should display the add rows button by default', () => {
+        const addRowContainer = de.nativeElement.querySelector('.dot-add-rows__container');
+
+        fixture.detectChanges();
+
+        expect(addRowContainer.classList.contains('dot-add-rows__add')).toEqual(true);
+    });
+
+    it('should display a row after click on AddRows button', () => {
```
fmontes: This test will pass without the ```.click``` because even when the list it's "hidden" for the position the selector of the ```<li>``` will still be able to grab them. For this case it's better to check if the class change or the position of the top.

**NOTE:** When you are doing tests is a good practice to make them fail and them fix them to make them pass.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    it('should escape to previous step', () => {
+        comp.selectColumnState();
+
+        testHotKeysMock.callback(['esc']);
+
+        expect(comp.rowState).toBe('add');
+    });
+
+    it('should be able to use left keyboard to highlight columns', () => {
```
fmontes: In this case you are valitading that the use of the key left never goes less than 0, thats ok to validate but the titleof the test doesn't sad that.

Same with the next test.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,122 @@
...
+        this.selectedLi = col;
+        event.stopPropagation();
+    }
+
+    /**
+     * Emit number of columns after select colum the reset row
+     * @memberof ContentTypeFieldsAddRowComponent
+     */
+    selectRow(): void {
+        this.selectColums.emit(this.selectedLi + 1);
```
fmontes: The method is named "selectoRow" but emit columns? that's weird.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    it('should select columns number after click on li', () => {
+        fixture.detectChanges();
+        const lis = de.queryAll(By.css('li'));
+
+        spyOn(comp.selectColums, 'emit');
+        comp.selectColums.subscribe(col => {
+            expect(col).toEqual(2);
+        });
+
+        lis[1].nativeElement.click(2);
```
fmontes: Why are you passing a 2 to the click? click method doesn't have params: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click

I don't think this test it's testing what you expect, I change the 2 for a 3 in the click and still pass, you need to revisit this.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,124 @@
+import { ContentTypeFieldsAddRowComponent } from './content-type-fields-add-row.component';
```
fmontes: You are missing a scenario here, you need to test that the correct <li> is active when you hit left or right in the keyboard...

Like if you are in ```li[0]``` and you hit right ```li[0]``` lost the focus and ```li[1]``` get the focus.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.scss
```diff
@@ -0,0 +1,83 @@
...
+:host {
+    display: flex;
+    flex-grow: 1;
+    flex-direction: column;
+    margin-bottom: $basic-padding;
+    padding: 0 $basic-padding * 2;
+    max-height: $row-height;
+}
+
+.dot-add-rows__container {
```
fmontes: You don't need to nest this classes, another of the benefits of use BEM is that the specifity comes "for free" because the name of the classes, it's kind of implicit.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.scss
```diff
@@ -0,0 +1,83 @@
...
+        width: 100%;
+        transition: top 200ms ease-in-out, bottom 200ms ease-out;
+    }
+
+    .dot-add-rows-button__container {
+        text-align: center;
+        height: $row-height;
+        align-items: center;
+        justify-content: center;
+        top: -70px;
```
fmontes: You can do ```top: -#{$row-height}```, check this sass interpolation variables for more info: http://krasimirtsonev.com/blog/article/SASS-interpolation-in-a-name-of-variable-nest-variables-within-variables

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+        });
+
+        fixture = DOTTestBed.createComponent(ContentTypeFieldsAddRowComponent);
+        de = fixture.debugElement;
+        comp = fixture.componentInstance;
+        el = de.nativeElement;
+    });
+
+    it('should render disabled input', () => {
+        disabled = true;
```
fmontes: You don't need this var here, you can assert directly to ```true/false```.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+        comp.disabled = disabled;
+
+        fixture.detectChanges();
+
+        const buttonElement = de.query(By.css('button'));
+
+        expect(comp.disabled).toEqual(buttonElement.nativeElement.disabled);
```
fmontes: So in the expects we should always do:

```expect(THING_WE_ARE_TESTING).to(OUR_TEST_VALUE)```

So in this case:

```expect(buttonElement.nativeElement.disabled).toEqual(true);``` <-- because you know you are setting the ```true```.

File: src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts [hunk 27]
```diff
@@ -0,0 +1,269 @@
...
+    });
+
+    it('should render columns input', () => {
+        const columnSelectionList = de.query(By.css('.dot-add-rows-columns-list__container'));
+        columns = [1, 2, 3];
```
fmontes: Don't need to store a var to pass params to the components, I don't see the value on it, maybe am missing something?

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+
+        comp.columns = columns;
+
+        fixture.detectChanges();
+
+        expect(comp.columns.length).toEqual(columnSelectionList.children.length);
```
fmontes: ```expect(THING_WE_ARE_TESTING).to(OUR_TEST_VALUE)```

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+    });
+
+    it('should render tooltip input data', () => {
+        toolTips = ['contenttypes.content.one_column',
```
fmontes: Same thing with the vars.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+        'contenttypes.content.four_columns'];
+
+        comp.ngOnInit();
```
fmontes: ```fixture.detectChanges();``` will trigger ```ngOnInit()``` not 100% sure but I think in this case it's being trigger twice.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+        comp.toolTips = toolTips;
+
+        fixture.detectChanges();
+
+        const columnSelectionList = de.query(By.css('.dot-add-rows-columns-list__container'));
+        const columnSelectionItems = columnSelectionList.nativeElement.children;
+
+        expect(comp.i18nMessages[toolTips[1]]).toEqual(columnSelectionItems[1].attributes[9].textContent);
```
fmontes: ```expect(THING_WE_ARE_TESTING).to(OUR_TEST_VALUE)```

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+        const addRowContainer = de.query(By.css('.dot-add-rows-columns-list__container'));
+
+        const addButton = de.query(By.css('button'));
+        addButton.nativeElement.click();
+        fixture.detectChanges();
+
+        expect(addRowContainer.nativeElement.classList.contains('dot-add-rows__select')).toEqual(true);
+    });
+
+   it('should focus the first column selection after click on Add Rows button', (done) => {
```
fmontes: What's this ```done``` param here for?

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+        fixture.detectChanges();
+
+        comp.selectColumnState();
+
+        const columnSelectionList = de.query(By.css('.dot-add-rows-columns-list__container'));
+        const firstElement = columnSelectionList.children[0];
+
+        setTimeout(() => {
```
fmontes: Why do we need a timeout here?

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /**
+     * Display row when click on add rows button
+     * @memberof ContentTypeFieldsAddRowComponent
+     */
+    selectColumnState(): void {
+        this.rowState = 'select';
+        this.bindKeyboardEvents();
+        setTimeout(() => { this.setFocus(this.getElementSelected()); }, 400);
```
fmontes: Not sure what's going need with this, maybe you can show me.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,188 @@
...
+    constructor(private hotkeysService: HotkeysService , public messageService: MessageService) {
+    }
+
+    ngOnInit(): void {
+        this.setKeyboardEvent('ctrl+a', this.selectColumnState.bind(this));
+        this.loadMessages();
+    }
+
+    ngOnDestroy(): void {
+        this.hotkeysService.remove(this.hotkeysService.get(['left', 'right', 'enter', 'esc']));
```
fmontes: What happen here is we try to "remove" hotkeys that doesn't exist? Because the component can be destroy and the keys are not always bind it.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.ts
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        this.setFocus(this.getElementSelected());
+    }
+
+    private resetState(): any {
+        this.removeFocus(this.getElementSelected());
+        this.selectedColumnIndex = 0;
+        this.rowState = 'add';
+        this.hotkeysService.remove(this.hotkeysService.get(['left', 'right', 'enter', 'esc']));
```
fmontes: We are doing this exact same thing on the ```ngOnDestroy``` move this to a private method.

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
freddyucv: Yes, it's being tigger twice

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
freddyucv: use fakeAsync and tick instead

src/app/test/hotkeys-service.mock.ts
```diff
@@ -0,0 +1,46 @@
...
+    remove(hotkey?: Hotkey | Hotkey[]): Hotkey | Hotkey[] {
+        return null;
+    }
+
+    callback(combo: string[]): any|void {
+        const hotkey: any = this.get(combo);
+        if (hotkey) {
+          return hotkey.callback(null, combo);
+        }
+    }
```
freddyucv: the methods need doc

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
luisrojascr: I used done param to include done() inside the setTimeout function. I'll use fakeAsync and tick.

luisrojascr: FakeAsync didn't work, I needed to add the setTimeout function in order to fix the focus issue  with the transition effect. https://stackoverflow.com/questions/10955201/how-to-test-a-function-which-has-a-settimeout-with-jasmine  

This is the TS code:
selectColumnState(): void {
        this.rowState = 'select';
        this.bindKeyboardEvents();
        setTimeout(() => { this.setFocus(this.getElementSelected()); }, 400);
    }

src/app/portlets/content-types/fields/content-type-fields-add-row/content-type-fields-add-row.component.spec.ts
luisrojascr: I needed to add the setTimeout function in order to fix the focus issue with the transition effect. https://stackoverflow.com/questions/10955201/how-to-test-a-function-which-has-a-settimeout-with-jasmine

This is the TS code:

selectColumnState(): void {
        this.rowState = 'select';
        this.bindKeyboardEvents();
        setTimeout(() => { this.setFocus(this.getElementSelected()); }, 400);
    }

fmontes: @luisrojascr show me this and turns out that transition doesn't work as expected when one of the transitioned elements it's focused.

PR #449, opened 2017-10-21
Issue 12735 save url when user not login

hmoreras:


src/app/view/components/login/login-component/login-container.component.ts
```diff
@@ -28,6 +29,7 @@ export class LoginContainerComponent {
 
     constructor(
         private dotRouterService: DotRouterService,
+        private dotNavigationService: DotNavigationService,
```
hmoreras: The include of this service here is needed to bind the events on User/Menu changes.

src/app/view/components/dot-navigation/dot-navigation.service.ts
```diff
@@ -18,11 +19,16 @@ export class DotNavigationService {
...
         private dotcmsEventsService: DotcmsEventsService,
         private loginService: LoginService,
-        private location: PlatformLocation
+        private location: PlatformLocation,
+        private router: Router
     ) {
-        this.dotMenuService.loadMenu().subscribe((menu: DotMenu[]) => {
-            this.setMenu(menu);
-        });
+        router.events
```
hmoreras: Load menu when Route Navigation ends if is a private page

src/app/api/services/guards/auth-guard.service.ts
```diff
@@ -11,10 +11,11 @@ import { DotRouterService } from './../dot-router-service';
 export class AuthGuardService implements CanActivate {
     constructor(private dotRouterService: DotRouterService, private loginService: LoginService) {}
 
-    canActivate(): Observable<boolean> {
+    canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
```
fmontes: Do we need to pass both params if we are not using them?

src/app/view/components/dot-navigation/dot-navigation.service.ts
fmontes: I'm not sure about this but, doesn't this will trigger every time we navigate to one place to another? and also, why do we need this ```NavigationEnd``` anyways?

src/app/view/components/dot-navigation/dot-navigation.service.ts
```diff
@@ -85,14 +92,20 @@ export class DotNavigationService {
...
+                                .gotoPortlet(this.dotRouterService.previousSavedURL, true)
+                                .then(res => {
+                                    // Menu update happens on NavigationEnd event
+                                    this.dotRouterService.previousSavedURL = null;
+                                });
+                        } else {
+                            this.goToFirstPortlet(); // Menu update happens on NavigationEnd event
+                        }
                     } else {
                         this.setMenu(menu);
```
fmontes: But also menu it's updating here...

src/app/view/components/login/login-component/login-container.component.ts
fmontes: Include this in this component just because we need to instace it before we use it doesn't seems right to me... If I'm doing a clean up unused code I'll certainly remove this because in a few months I will forgot why it was there.

We need to come up with a better solution for it. @freddyucv ideas?

src/app/api/services/guards/auth-guard.service.ts
hmoreras: Yes, because we are using the the second one, state.

src/app/view/components/dot-navigation/dot-navigation.service.ts
hmoreras: The subscription to the NavigationEnd, is needed, because we need to update the menu item once the route is served and when non public page. In the previous scenario we were asking for a menu update before the route was updated, and that was causing issues with the selected option. 

Regarding the menu update you mention in this comment, is only executed when we use the functionality of "Login as" or we add or remove a new item in the menu. So is not duplicate functionality, each serve in different scenarios.

src/app/view/components/dot-navigation/dot-navigation.service.ts
hmoreras: The subscription to the NavigationEnd, is needed, because we need to update the menu item once the route is served and when non public page. In the previous scenario we were asking for a menu update before the route was updated, and that was causing issues with the selected option.

src/app/view/components/dot-navigation/dot-navigation.service.ts
```diff
@@ -9,6 +9,7 @@ import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 import { DotcmsEventsService } from 'dotcms-js/core/dotcms-events.service';
 import { Auth } from 'dotcms-js/core/login.service';
 import { Router, NavigationEnd } from '@angular/router';
+import set = Reflect.set;
```
fmontes: What is this?

src/app/view/components/dot-navigation/dot-navigation.service.ts
```diff
@@ -100,9 +101,11 @@ export class DotNavigationService {
                         if (this.dotRouterService.previousSavedURL) {
                             this.dotRouterService
                                 .gotoPortlet(this.dotRouterService.previousSavedURL, true)
-                                .then(res => {
-                                    this.dotRouterService.previousSavedURL = null;
-                                    this.setMenu(menu);
+                                .then((res) => {
```
fmontes: ```(res: boolean)```

PR #450, opened 2017-10-25
Issue 12782 loading indicator not showing

hmoreras:


src/app/view/components/login/login-component/login-container.component.ts
```diff
@@ -30,8 +31,10 @@ export class LoginContainerComponent {
         private dotRouterService: DotRouterService,
         private httprequestUtils: HttpRequestUtils,
         private loggerService: LoggerService,
-        private loginService: LoginService
+        private loginService: LoginService,
+        private dotLoadingIndicatorService: DotLoadingIndicatorService
     ) {
+        this.dotLoadingIndicatorService.hide();
```
fmontes: Why do you need it right here? by default the loading indicator it hidden, right? if not I think we have to make it hidden by default.

src/app/view/components/login/login-component/login-container.component.ts
```diff
@@ -44,30 +47,24 @@ export class LoginContainerComponent {
 
     logInUser(loginData: LoginData): void {
         this.isLoginInProgress = true;
```
fmontes: This need to be clean up because it's not being used now.

src/app/view/components/login/login-component/login.component.ts
```diff
@@ -71,10 +69,12 @@ class LoginComponent {
     constructor(
         private loginService: LoginService,
         private ngZone: NgZone,
-        private loggerService: LoggerService
+        private loggerService: LoggerService,
+        private dotLoadingIndicatorService: DotLoadingIndicatorService
     ) {
         this.language = '';
         this.renderPageData();
+        dotLoadingIndicatorService.hide();
```
freddyucv: The same before Why do you need it right here?

src/app/view/components/login/login-component/login.component.ts
```diff
@@ -111,6 +111,7 @@ class LoginComponent {
                 error += this.mandatoryFieldError.replace('{0}', this.passwordLabel);
             }
             this.message = error;
+            this.dotLoadingIndicatorService.hide();
```
freddyucv: The same before Why do you need it right here?, you are showing or hiding the loading indicator in LoginContainer so I think you dont need it in this Component

src/app/view/components/login/login-component/login-container.component.ts
hmoreras: this is still used, to disable input fields while doing the login.

PR #453, opened 2017-10-30
Issue 12728 back button not updating url

hmoreras:


src/app/view/components/_common/iframe/iframe-porlet-legacy/iframe-porlet-legacy.component.html
```diff
@@ -1 +1 @@
-<dot-iframe #iframe [src]="url | async"></dot-iframe>
+<dot-iframe #iframe [src]="url | async" [isLoading]="isLoading"></dot-iframe>
```
fmontes: I discuss this with @hmoreras and we think that for now it's the best solution, because the other solution it's to manually manage the ```history state``` of the browser and it could be really messy.

src/app/view/components/_common/iframe/iframe-porlet-legacy/iframe-porlet-legacy.component.ts
```diff
@@ -106,30 +106,23 @@ export class IframePortletLegacyComponent implements OnInit {
...
-                    ? this.contentletService.getUrlById(id)
-                    : this.dotNavigationService.getUrlById(id);
+                return url === 'add' ? this.contentletService.getUrlById(id) : this.dotNavigationService.getUrlById(id);
             })
             .subscribe((url: string) => {
                 this.setUrl(url);
             });
     }
+
+    /**
```
fmontes: Private methods shouldn't have doc

jdotcms: In general private method does not need to be documented, however documentation is good!

hmoreras: I prefer to keep a comment on this one, to remember why we are removing the iframe from the DOM with the ngIf.

PR #458, opened 2017-10-31
dotCMS/core#12876 Content type edit with unknown id breaks the app

fmontes:


src/app/portlets/content-types/content-types-resolver.service.ts
```diff
@@ -42,6 +42,10 @@ export class ContentTypeResolver implements Resolve<ContentType> {
                     this.dotRouterService.gotoPortlet('/content-types-angular');
                     return null;
                 }
+            })
+            .catch(() => {
+                this.dotRouterService.gotoPortlet('/content-types-angular');
```
hmoreras: here, use as a second parameter, 'true'. so the URL is replaced and don't keep the invalid entry in the browser history

fmontes: Good catch!

PR #460, opened 2017-11-02
Issue 12781 cant add key values or vanity url

hmoreras:


src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -24,133 +26,107 @@ import { MessageService } from '../../../api/services/messages-service';
...
-                            this.createContentType('widget', $event);
-                        },
-                        icon: 'fa-cog',
-                        label: this.i18nMessages['contenttypes.content.widget']
+    ngOnInit() {
+        Observable.forkJoin(
+            this.messageService.getMessages(this.i18nKeys),
+            this.dotContentletService.getContentTypes()
+        ).subscribe(val => {
+            this.i18nMessages = val[0];
```
fmontes: I think you are overwriting the ```i18nMessages``` prop in the BaseComponent not sure it's that's right, please double check.

src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -24,133 +26,107 @@ import { MessageService } from '../../../api/services/messages-service';
...
-                    {
-                        command: $event => {
-                            this.createContentType('widget', $event);
-                        },
-                        icon: 'fa-cog',
-                        label: this.i18nMessages['contenttypes.content.widget']
+    ngOnInit() {
+        Observable.forkJoin(
+            this.messageService.getMessages(this.i18nKeys),
+            this.dotContentletService.getContentTypes()
```
fmontes: This response with a ```StructureTypeView[]``` make sure you use data types.

src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -24,133 +26,107 @@ import { MessageService } from '../../../api/services/messages-service';
...
-                    {
-                        command: $event => {
-                            this.createContentType('file', $event);
-                        },
-                        icon: 'fa-file-o',
-                        label: this.i18nMessages['contenttypes.content.file']
+                    model: []
+                }
+            };
+            val[1].filter(r => ['RECENT_CONTENT', 'RECENT_WIDGET'].indexOf(r.name) === -1).forEach(obj => {
```
fmontes: Extract these functions to external method to make this more readable.

Also, you can use ```.map``` instead of ```.forEach``` and don't do ```.push``` but just assign the whole thing becahse ```.map``` return an array.

src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -24,133 +26,107 @@ import { MessageService } from '../../../api/services/messages-service';
...
+                this.actionHeaderOptions.primary.model.push({
+                    command: $event => {
```
fmontes: Do you need the anonymous function here? can't you just pass the method?

src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -24,133 +26,107 @@ import { MessageService } from '../../../api/services/messages-service';
...
-                label: 'Remove',
-                icon: 'fa-trash',
-                command: (item) => this.removeConfirmation(item)
-            }
-        ];
+                    icon: this.contentTypesInfoService.getIcon(obj.name),
+                    label: this.i18nMessages['contenttypes.content.' + obj.name.toLocaleLowerCase()]
+                });
+            });
+            this.contentTypeColumns = [
```
fmontes: Move this to a external function so ```ngOnInit``` it's not that big and @freddyucv don't frak out, like 

```this.contentTypeColumns = this.getTypeColumns()```

File: src/app/portlets/content-types/main/content-types.component.ts [hunk 6]
```diff
@@ -4,13 +4,15 @@ import { DotConfirmationService } from './../../../api/services/dot-confirmation
 import { CrudService } from './../../../api/services/crud';
 import { MenuItem } from 'primeng/primeng';
 import { ActivatedRoute, Router } from '@angular/router';
-import { Component, ViewChild } from '@angular/core';
+import { Component, OnInit, ViewChild } from '@angular/core';
 
 import { ActionHeaderOptions } from '../../../shared/models/action-header';
 import { BaseComponent } from '../../../view/components/_common/_base/base-component';
```
luisrojascr: Remove BaseComponent import

src/app/portlets/content-types/main/content-types.component.ts
hmoreras: BaseComponent is not longer used here.

src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -1,119 +1,110 @@
...
+            this.messageService.getMessages(this.i18nKeys),
+            this.dotContentletService.getAllContentTypes()
+        ).subscribe(val => {
+            this.i18nMessages = val[0];
+            this.actionHeaderOptions = {
+                primary: {
+                    command: $event => {
+                        this.createContentType($event);
+                    },
+                    model: this.setContentTypes(val[1])
```
fmontes: ```val[1]``` is an ```StructureTypeView[]``` please create a const with a more semantic name and use the data type.

src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -1,119 +1,110 @@
...
     }
 
-    /**
-     * Callback call from BaseComponent when the messages are received.
-     * @memberOf ContentTypesPortletComponent
-     */
-    onMessage(): void {
-        this.actionHeaderOptions = {
-            primary: {
+    private setContentTypes(arr: StructureTypeView[]): ButtonModel[] {
```
fmontes: Please use sematyc name, ```arr``` could be ```baseTypes```

(I know the data datatype is ```StructureTypeView[]``` but we have to change that interface name, it's a old code```

src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -1,119 +1,110 @@
...
+        private dotConfirmationService: DotConfirmationService,
+        private dotContentletService: DotContentletService
+    ) {}
+
+    ngOnInit() {
+        Observable.forkJoin(
+            this.messageService.getMessages(this.i18nKeys),
+            this.dotContentletService.getAllContentTypes()
+        ).subscribe(val => {
+            this.i18nMessages = val[0];
```
fmontes: We can't save this here, I like the idea of not using the ```BaseComponent``` here, but save the messages in every component where we are going to use it it's not a good idea and it's not leaving in a better position of where we want to go... when you ask the the ```messageService``` for messages they are store in the ```messageService```, here: https://github.com/dotCMS/core-web/blob/master/src/app/api/services/messages-service.ts#L111

So what we need to do it's create a method in the message service like:

```
get(key: string): Observable<string> {
    return this.messagesLoaded.filter(SOMETHING HERE)
}
```

I think something like this could work... but we need to test.

PR #463, opened 2017-11-06
Issue 12983 hide delete button in undeleteable

hmoreras:


src/app/view/components/_common/action-menu-button/action-menu-button.component.spec.ts
```diff
@@ -97,7 +109,40 @@ describe('ActionMenuButtonComponent', () => {
...
+                },
+                shouldShow: () => true
+            }
+        ];
+
+        comp.actions = fakeActions;
+        comp.ngOnInit();
+
+        expect(comp.filteredActions.length).toEqual(1);
+    });
```
freddyucv: I think we need a test where shouldShow return false just for some items and true for the others

hmoreras: that is done in this test, I am asking if filteredActions is equals to 1, when the actions array is 3.

PR #466, opened 2017-11-13
dotCMS/core#13005

luisrojascr:


src/app/view/components/_common/searchable-dropdown/component/searchable-dropdown.component.ts
```diff
@@ -45,6 +45,7 @@ export class SearchableDropdownComponent extends BaseComponent implements Contro
     @Input() rows: number;
     @Input() totalRecords: number;
     @Input() placeholder = '';
+    @Input() persistentPlaceholder: string;
```
freddyucv: persistentPlaceholder should be a boolean, if this is true then placeholder is always showed.

Template:

[label]="persistentPlaceholder ? placeholder : valueString"

src/app/view/components/container-selector/container-selector.component.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+        fixture.detectChanges();
+
+        const paginatorService: PaginatorService = de.injector.get(PaginatorService);
+        paginatorService.totalRecords = 2;
+        spyOn(paginatorService, 'getWithOffset').and.returnValue(Observable.of([]));
+
+        fixture.detectChanges();
+
+        this.searchableDropdownComponent.filterChange.emit(filter);
+        comp.handleFilterChange(filter);
```
freddyucv: You should not need call the handleFilterChange method here because is it the callback for 'filterChange.emit'

src/app/view/components/container-selector/container-selector.component.ts
```diff
@@ -0,0 +1,76 @@
...
+    currentContainers: Container[] = [];
+    selectedContainersList: Container[] = [];
+
+    constructor(public paginationService: PaginatorService, public messageService: MessageService) { }
+
+    ngOnInit(): void {
+        this.paginationService.url = 'v1/containers';
+        this.messageService.getMessages([
+            'page.container.add-container'
+        ]).subscribe();
```
freddyucv: We have to do something with this fast

src/app/view/components/container-selector/container-selector.component.html
```diff
@@ -0,0 +1,23 @@
+<div class="container-selector">
+    <div class="head-container">
+        <searchable-dropdown
+            (change)="containerChange($event)"
+            (filterChange)="handleFilterChange($event)"
+            (pageChange)="handlePageChange($event)"
+            [data]="currentContainers"
+            [pageLinkSize]="3"
```
fmontes: By default it's 3, les pass a 5 here.

src/app/view/components/container-selector/container-selector.component.html
```diff
@@ -0,0 +1,23 @@
+<div class="container-selector">
+    <div class="head-container">
```
fmontes: ```.container-selector__header```

src/app/view/components/container-selector/container-selector.component.html
```diff
@@ -0,0 +1,23 @@
...
+            (pageChange)="handlePageChange($event)"
+            [data]="currentContainers"
+            [pageLinkSize]="3"
+            [rows]="paginationService.paginationPerPage"
+            [totalRecords]="paginationService.totalRecords"
+            class="container-selector"
+            labelPropertyName="name"
+            persistentPlaceholder="{{messageService.get('page.container.add-container')}}">
+        </searchable-dropdown>
+        <button class="head-container__close-containers" pButton iconButton flat icon="fa-close" (click)="remove.emit($event)"></button>
```
fmontes: ```.container-selector__header-remove```

src/app/view/components/container-selector/container-selector.component.scss
```diff
@@ -0,0 +1,54 @@
...
+    border: 1px solid $gray-lighter; 
+}
+
+.head-container {
+    height: 48px; 
+    padding: $basic-padding $basic-padding * 2; 
+    background-color: $gray-lighter; 
+
+    .ui-button {
+        background: transparent;
```
fmontes: I don't think you need this if you use a flat button.

src/app/view/components/container-selector/container-selector.component.scss
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+.head-container__close-containers {
+    float: right;
+    padding-top: $basic-padding;
```
fmontes: Use flexbox... so you can dynamically vertical center this no matter what the height is... and use flexbox always if you can.

src/app/view/components/container-selector/container-selector.component.scss
```diff
@@ -0,0 +1,54 @@
...
+    cursor: pointer;
```
fmontes: Doesn't button have this by default?

src/app/view/components/container-selector/container-selector.component.scss
```diff
@@ -0,0 +1,54 @@
...
+    color: $gray;
+
+    &:hover {
+        color: $black;
```
fmontes: When a text or an icon use the color of the body text like in this case, use ```$body-font-color```

src/app/view/components/container-selector/container-selector.component.scss
```diff
@@ -0,0 +1,54 @@
...
+        border-top: 1px solid $gray-lighter;
+        padding: $basic-padding 0;
+
+        &:first-child {
+            border-top: none;
+        }
+    }
+}
+
+.ui-button[iconButton] {
```
fmontes: We need to do this change globally... not sure what this looks like tho.

src/app/view/components/container-selector/container-selector.component.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+                PaginatorService
+            ]
+        });
+
+        fixture = DOTTestBed.createComponent(ContainerSelectorComponent);
+        comp = fixture.componentInstance;
+        de = fixture.debugElement;
+        el = de.nativeElement;
+
+        this.searchableDropdownComponent = de.query(By.css('searchable-dropdown')).componentInstance;
```
fmontes: ```let searchableDropdownComponent```
and
```let containers```

Just like we do with comp, fixture... etc.

src/app/view/components/container-selector/container-selector.component.scss
luisrojascr: Actually that's for searchable-dropdown button to match the mock.

fmontes: I run the project to see where is is button and I wansn't able to find it... can you point it to me?

PR #474, opened 2017-11-29
dotCMS/core#13102 Style layout component

fmontes:


src/app/app.component.html
```diff
@@ -1,7 +1,7 @@
 <router-outlet></router-outlet>
 <p-confirmDialog width="400" #cd>
     <p-footer>
-        <button type="button" class="button-delete" pButton [label]="(dotConfirmationService.labels | async)?.acceptLabel" (click)="cd.accept()"></button>
-        <button type="button" pButton [label]="(dotConfirmationService.labels | async)?.rejectLabel" (click)="cd.reject()"></button>
+        <button type="button" pButton flat [label]="(dotConfirmationService.labels | async)?.rejectLabel" (click)="cd.reject()"></button>
+        <button type="button" pButton flat [label]="(dotConfirmationService.labels | async)?.acceptLabel" (click)="cd.accept()"></button>
```
hmoreras: looks like flat is just used to style up the buttons, so instead to be an attribute I would say is a good practice to move it to a class.

fmontes: Agreed... I'll probably do this in the refactor of the buttons I told you guys about at stand up.

PR #477, opened 2017-11-30
Implement global event handler service

hmoreras:


src/app/shared/models/event/dot-event.ts
```diff
@@ -0,0 +1,9 @@
+/**
```
fmontes: Name the folder ```dot-event``` too because event it's way too common.

src/app/view/components/dot-toolbar/dot-toolbar.component.ts
```diff
@@ -1,24 +1,30 @@
...
+        private dotEventsService: DotEventsService
+    ) {}
 
     siteChange(site: Site): void {
         this.siteService.switchSite(site);
     }
 
     handleMainButtonClick($event): void {
         this.mainButtonClick.emit($event);
+        this.dotEventsService.notify({name: 'dot-toolbar-toggle'});
```
fmontes: This should go here: https://github.com/dotCMS/core-web/blob/master/src/app/view/components/main-legacy/main-legacy.component.ts#L29 because that's the place where we do toggle the side navigation... and the event should be named like: ```dot-side-nav-toogle``` because it's related with show/hide the side nav, no matter which button we toogle it with or where is located.

src/app/view/components/dot-toolbar/dot-toolbar.component.ts
```diff
@@ -1,24 +1,30 @@
...
 import { SiteService, Site } from 'dotcms-js/dotcms-js';
 import { IframeOverlayService } from '../_common/iframe/service/iframe-overlay.service';
+import { DotEventsService } from '../../../api/services/dot-events.service';
 
 @Component({
     encapsulation: ViewEncapsulation.None,
-    selector: 'toolbar-component',
-    styleUrls: ['./toolbar.component.scss'],
-    templateUrl: './toolbar.component.html'
+    selector: 'dot-toolbar-component',
```
fmontes: Thanks! but from the selector remove the ```-component``` should be ```<dot-toolbar>```

src/app/portlets/dot-edit-page/layout/dot-edit-layout-grid/dot-edit-layout-grid.component.ts
```diff
@@ -70,12 +72,18 @@ export class DotEditLayoutGridComponent implements OnInit, ControlValueAccessor
...
+        private dotEventsService: DotEventsService
     ) {}
 
     ngOnInit() {
         this.messageService.getMessages(this.i18nKeys).subscribe();
         this.setGridValue();
+        this.dotEventsService.listen('dot-toolbar-toggle').subscribe(() => {
+            setTimeout(() => {
+                this.ngGrid.triggerResize();
+            }, 150);
```
fmontes: Plase add a comment here where you said why is this 150 timeout for.

src/app/shared/models/event/dot-event.ts
```diff
@@ -0,0 +1,9 @@
...
+ * Interface for custom events.
+ *
+ * @interface
+ */
+export interface DotEvent {
+    name: string;
+    data?: any;
```
jdotcms: this is usually called payload but I guess data is ok

src/app/api/services/dot-events.service.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Observable } from 'rxjs/Observable';
+import { Subject } from 'rxjs/Subject';
+import { DotEvent } from '../../shared/models/event/dot-event';
+
+/**
+ * Provide a Global service to Subscribe to custom events and notify subscribers when those events occur.
+ * @DotEventsService
+ */
+@Injectable()
+export class DotEventsService {
```
jdotcms: I am not aware about the intention here and I am not an expert on TS, neither angular; however it seems your Service only has the ability to subscribe one subscriber. If it is a singleton over the app I would give the ability to add more than one listener.
However and as I said, I am not an angular guy so not sure if my comment is out of context and I am wrong

hmoreras: The `BehaviorSubject ` overwrite the current value with a new one each time the `notify` method is called, and all the subscribers are listening to the `BehaviorSubject ` in a specific event type. In general has the ability to subscribe and react as many events as required.

src/app/api/services/dot-events.service.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    it('should notify subscribers', () => {
+        let numbersArray: number[] = [];
+        dotEventsService.listen('test').subscribe(value => {
+            numbersArray = value.data;
+        });
+
+        dotEventsService.notify(testEvent);
+
+        expect(numbersArray.length).toEqual(3);
```
freddyucv: you should test:

expect(numbersArray).toEqual([1,2,3]);

not just the length

src/app/portlets/dot-edit-page/layout/dot-edit-layout-grid/dot-edit-layout-grid.component.spec.ts
```diff
@@ -48,6 +49,7 @@ describe('DotEditLayoutGridComponent', () => {
                 DotConfirmationService,
                 DotEditLayoutService,
                 PaginatorService,
+                DotEventsService,
```
freddyucv: do you need add a new test where this service is used?, you have to create a mock of EventService and tigger the listen method for 'dot-side-nav-toggle' event and then check that the this.ngGrid.triggerResize method is called

src/app/portlets/dot-edit-page/layout/dot-edit-layout/dot-edit-layout.component.spec.ts
```diff
@@ -66,6 +67,7 @@ describe('DotEditLayoutComponent', () => {
                 PaginatorService,
                 SocketFactory,
                 DotEditLayoutService,
+                DotEventsService,
```
freddyucv: you must add a new test here to test the DptEventService used

hmoreras: This is not used in this component at all, just added because the a child use it, actually I think the provider of this service need to go in the DOTTestBed since is global.

PR #490, opened 2017-12-13
Issue 12936 integrate workflow end point

hmoreras:


src/app/api/services/workflow/workflow.service.ts
```diff
@@ -12,7 +12,13 @@ import {Workflow} from '../../../shared/models/workflow/workflow.model';
  */
 @Injectable()
 export class WorkflowService {
-    constructor(private coreWebService: CoreWebService) {}
+    private urls: any;
+
+    constructor(private coreWebService: CoreWebService) {
+        this.urls = {
+            getWorkflowSchemes: 'v1/workflow/schemes'
```
fmontes: We do this if the service needs to handle more that one url... in this case you can just put it directly where you define the property in line 15.

src/app/api/services/workflow/workflow.service.ts
```diff
@@ -21,28 +27,11 @@ export class WorkflowService {
...
-                identifier: '4'
-            },
-            {
-                name: 'WorkFlow #5',
-                identifier: '5'
-            }
-        ];
-        return Observable.of(dummyWorkflows);
+        return this.coreWebService
+            .request({
```
fmontes: Use ```requestView``` method... request is old, please don't use it. cc: @luisrojascr

PR #491, opened 2017-12-14
Issue 13178 stopped sites need to be available in site selector

luisrojascr:


src/app/view/components/_common/site-selector/site-selector.component.html
```diff
@@ -16,4 +16,4 @@
     >
 </searchable-dropdown>
 
-<span *ngIf="totalRecords == 1" class="site-selector__title">{{currentSite && currentSite.value ? currentSite.value.hostname : '' | async}}</span>
\ No newline at end of file
+<span *ngIf="totalRecords == 1" class="site-selector__title">{{currentSite && currentSite['value'] ? currentSite['value'].hostname : '' | async}}</span>
```
luisrojascr: When doing this 'currentSite.value' it had a warning "Identifier value is not defined. "

PR #492, opened 2017-12-17
Issue 13215 update sidenav for ipad and tables

luisrojascr:


src/app/view/components/dot-navigation/dot-navigation.component.ts
```diff
@@ -12,13 +12,30 @@ import { DotNavigationService } from './dot-navigation.service';
...
 
+    /**
+     * Respond to document events and collapse the sidenav if is clicked outside
+     * TODO: not working on iframes
+     * @param {*} event
+     * @memberof DotNavigationComponent
+     */
+    @HostListener('document:click', ['$event'])
+    handleClick(event: any) {
+        if (!this.el.nativeElement.contains(event.target)) {
```
fmontes: This condition it's too complicated, double check it and move this to a method with a name that tells what is this condition about.

src/app/view/components/dot-navigation/dot-navigation.component.ts
```diff
@@ -12,13 +12,30 @@ import { DotNavigationService } from './dot-navigation.service';
...
 
-    constructor(private dotNavigationService: DotNavigationService) {}
+    constructor(private dotNavigationService: DotNavigationService, private el: ElementRef) {}
 
     ngOnInit() {
         this.menu = this.dotNavigationService.items$;
     }
 
+    /**
+     * Respond to document events and collapse the sidenav if is clicked outside
```
fmontes: Check this service ```IframeOverlayService``` and the way we use it... I think this could be a good fit for this.

src/app/view/components/dot-toolbar/dot-toolbar.component.html
```diff
@@ -3,7 +3,7 @@
         <div class="ui-toolbar-group-left">
             <div class="toolbar__button-wrapper">
                 <div class="toolbar__button" [class.toolbar__button--hide]="collapsed">
-                    <button pButton icon="fa-arrow-left" class="inverted" (click)="handleMainButtonClick($event)"></button>
+                    <button pButton [icon]="icon" class="inverted" (click)="handleMainButtonClick($event)"></button>
```
fmontes: Sorry I wasn't clear enough but the change of the icon is for all screen sizes.

src/app/view/components/main-legacy/main-legacy.component.ts
```diff
@@ -18,6 +19,38 @@ export class MainComponentLegacy implements OnInit, OnDestroy {
...
+      event.target.innerWidth < 1025 ? this.isTablet = true : this.isTablet = false;
+    }
+
+    /**
+     * Set icon based on the window width
+     * @returns {string}
+     * @memberof MainComponentLegacy
+     */
+    setIcon(): string {
+        return window.innerWidth < 1025 ? 'fa-close' : 'fa-arrow-left';
```
fmontes: Sorry I wasn't clear enough but the change of the icon is for all screen sizes.

src/app/view/components/main-legacy/main-legacy.component.ts
```diff
@@ -18,6 +19,38 @@ export class MainComponentLegacy implements OnInit, OnDestroy {
...
+    }
+
+    /**
+     * Set isTablet when resizing the window size
+     * @param {any} event
+     * @memberof MainComponentLegacy
+     */
+    @HostListener('window:resize', ['$event'])
+        onResize(event: any) {
+      event.target.innerWidth < 1025 ? this.isTablet = true : this.isTablet = false;
```
fmontes:
```
this.isTablet = event.target.innerWidth < 1025
```

src/app/view/components/dot-navigation/dot-navigation.component.ts
```diff
@@ -12,22 +12,37 @@ import { DotNavigationService } from './dot-navigation.service';
 })
 export class DotNavigationComponent implements OnInit {
     menu: Observable<DotMenu[]>;
+    @Output() onMenuLinkClicked = new EventEmitter<Event>();
```
fmontes: Don't use "on" in the name of the events, check the angular styleguide: https://angular.io/guide/styleguide#dont-prefix-output-properties

src/app/view/components/dot-navigation/dot-navigation.component.ts
```diff
@@ -42,4 +57,10 @@ export class DotNavigationComponent implements OnInit {
     isActive(id: string) {
         return this.dotNavigationService.isActive(id);
     }
+
+    private isClickedOutsideSidenav(eventTarget: MouseEvent, isSafeToCollapse: boolean): void {
+        if (!this.sideNavEl.nativeElement.contains(eventTarget) && isSafeToCollapse) {
+                this.onMenuLinkClicked.emit(event);
```
fmontes: This logic doesn't make much sense, because:

If I click in the overlay the condition in the line 62 is ```true``` then an event called ```onMenuLinkClicked``` emits and I hide the menu because I'm listening to that, but I didn't click in a menu link like the event ```onMenuLinkClicked``` implied.

The event ```onMenuLinkClicked``` as implied should ONLY emit when I click on a menu link in this component.

But also, we don't need this if we do the comment before.

src/app/view/components/dot-navigation/dot-navigation.component.ts
```diff
@@ -12,22 +12,37 @@ import { DotNavigationService } from './dot-navigation.service';
...
     ngOnInit() {
         this.menu = this.dotNavigationService.items$;
     }
 
+    /**
+     * Respond to document events and collapse the sidenav if is clicked outside
+     * @param {*} event
+     * @memberof DotNavigationComponent
+     */
+    @HostListener('document:click', ['$event'])
```
fmontes: We should be listening to ```document:click``` in the ```main-legacy.component.ts``` because that's the one who can show and hide the sidenav, and there when a click happen in document we check the target... if the target is **not** the item that expand/collapse the accordion menus we hide the sidebar because it means the happen in something outside the sidenav OR in a sidenav link to navigate to another portlet.

src/app/view/components/main-legacy/main-legacy.component.scss
```diff
@@ -22,6 +24,49 @@ $sidebar-width: 190px;
 
 .layout__main {
     flex-grow: 1;
+
+    &.tabletOnly{
```
fmontes: BEM class names.

src/app/view/components/main-legacy/main-legacy.component.scss
```diff
@@ -22,6 +24,49 @@ $sidebar-width: 190px;
...
+            height: 100%;
+            top: 0;
+            bottom: 0;
+            left: 0;
+            right: 0;
+            opacity: 0.3;
+            z-index: 9;
+            animation: fadeIn 1s;
+    
+            @keyframes fadeIn {
```
fmontes: Why ```@keyframes``` instead of just ```transition```?

src/app/view/components/main-legacy/main-legacy.component.ts
```diff
@@ -18,6 +21,44 @@ export class MainComponentLegacy implements OnInit, OnDestroy {
...
+     * Respond to document events and collapse the sidenav if is clicked outside
+     * @param {*} event
+     * @memberof DotNavigationComponent
+     */
+    @HostListener('document:click', ['$event'])
+    onClickOutside(event: any) {
+        const eTarget = event.target;
+
+        // tslint:disable-next-line:max-line-length
+        if (!this.mainNav.sideNavEl.nativeElement.contains(eTarget) && eTarget.className !== 'layout__sidebar' && eTarget.className.split(' ')[1] !== 'ui-clickable') {
```
freddyucv: this 'this.mainNav.sideNavEl.nativeElement.contains(eTarget)' should be a mainNav method, something like: this.sideNavElContains(eTarget)

src/app/view/components/main-legacy/main-legacy.component.ts
```diff
@@ -18,26 +18,6 @@ export class MainComponentLegacy implements OnInit, OnDestroy {
...
-        this.router.events.subscribe(event => this.isTabletScreenOrLess());
-
-        this.isTabletScreenOrLess();
-    }
-
-    /**
-     * Set collapsed menu if window size is less than iPad landscape size
-     * @memberof MainComponentLegacy
-     */
-    isTabletScreenOrLess(): void {
```
fmontes: This name doesn't didn't describe what the method is doing.

src/app/view/components/main-legacy/main-legacy.component.ts
```diff
@@ -53,16 +33,41 @@ export class MainComponentLegacy implements OnInit, OnDestroy {
      * @param {*} event
      * @memberof DotNavigationComponent
      */
-    @HostListener('document:click', ['$event'])
+    @HostListener('click', ['$event'])
```
fmontes: Turns out that the document.click didn't work in the actual device (I tested in an old iPad 3 I have), so because the component is ALL the page we can listen to the click in the actual component and that worked.

src/app/view/components/main-legacy/main-legacy.component.ts
```diff
@@ -53,16 +33,41 @@ export class MainComponentLegacy implements OnInit, OnDestroy {
...
     onClickOutside(event: any) {
-        this.isMenuCollapsed = true;
+        if (this.isTablet && !this.isMenuCollapsed) {
```
fmontes: Well... just to make sure.

PR #494, opened 2017-12-20
Issue 13231 make dependencies local and encapsulate js code in edit content

freddyucv:


src/app/portlets/dot-edit-content/services/dot-edit-content-html.service.ts
```diff
@@ -296,7 +266,7 @@ export class DotEditContentHtmlService {
         this.addContainerToolbar();
         this.addContentletMarkup();
 
-        this.initDragAndDropContentlets();
+        this.dotDragDropAPIHtmlService.initDragAndDropContext(this.getEditPageDocument());
```
fmontes: Why instead of passing doc around we just don't make a method that return the <script> or whatever we need and we append it to the doc here? I think we should have the doc in just one place (the main service) if possible.

src/app/portlets/dot-edit-content/services/html/dot-drag-drop-api-html.service.ts
```diff
@@ -0,0 +1,44 @@
...
+        const dragulsJSElement = this.dotDOMHtmlUtilService.creatExternalScriptElement(
+            doc,
+            `${API_ROOT_PATH}/dragula.min.js`,
+            () => this.initDragula(doc)
+        );
+        doc.body.appendChild(dragulsJSElement);
+    }
+
+    private getDragAndDropCssElement(doc: any): any {
+        return ;
```
fmontes: ?

src/app/portlets/dot-edit-content/services/html/iframe-edit-mode.js.ts
```diff
@@ -0,0 +1,60 @@
...
+                    contentlet: {
+                        identifier: el.dataset.dotIdentifier,
+                        inode: el.dataset.dotInode
+                    }
+                }
+            });
+        }
+    })
+    // Init the model
+    window.model.next(getModel());
```
fmontes: Remember to rename this model var

src/app/portlets/dot-edit-content/services/dot-edit-content-html.service.ts
freddyucv: ... but you need the doc to create the element 

https://github.com/dotCMS/core-web/pull/494/files#diff-1661e25b47cd545049a58805e87cc482L246

PR #506, opened 2018-01-06
dotCMS/core#13270 Add dialog to choose edit template or layout

fmontes:


src/app/api/services/dot-messages-service.ts
```diff
@@ -42,7 +42,13 @@ export class DotMessageService {
     }
 
     get(key: string): string {
-        return this.messagesLoaded[key];
+        if (arguments.length > 1) {
```
freddyucv: it is not better use something like:

get(key: string, arguments: any[]): string

we have to add doc to this method too

src/app/portlets/dot-edit-page/layout/dot-edit-layout/dot-edit-layout.component.html
```diff
@@ -4,7 +4,7 @@
         <h3 class="dot-edit-layout__page-title" *ngIf="!saveAsTemplate">{{pageView.page.title}}</h3>
         <input #templateName pInputText type="text" placeholder="{{dotMessageService.get('editpage.layout.toolbar.template.name')}}" class="dot-edit-layout__toolbar-template-name" *ngIf="saveAsTemplate" formControlName="title">
         <div class="dot-edit-layout__toolbar-actions">
-            <p-checkbox class="dot-edit-layout__toolbar-save-template" label="{{dotMessageService.get('editpage.layout.toolbar.save.template')}}" *ngIf="isLayout()" binary="true" (onChange)="saveAsTemplateHandleChange($event)"></p-checkbox>
+            <p-checkbox class="dot-edit-layout__toolbar-save-template" label="{{dotMessageService.get('editpage.layout.toolbar.save.template')}}" *ngIf="isLayout() || form.get('title').value === null" binary="true" (onChange)="saveAsTemplateHandleChange($event)"></p-checkbox>
```
freddyucv: why do we need this condition?

'form.get('title').value === null'

src/app/portlets/dot-edit-page/layout/dot-edit-layout/dot-edit-layout.component.ts
```diff
@@ -19,11 +20,12 @@ export class DotEditLayoutComponent implements OnInit {
     @ViewChild('templateName') templateName: ElementRef;
 
     form: FormGroup;
-
     pageView: DotPageView;
     saveAsTemplate: boolean;
+    showDialogSelection = false;
```
freddyucv: I dont know if this is a good name for this attribute

src/app/portlets/dot-edit-page/layout/dot-edit-layout/dot-edit-layout.component.ts
```diff
@@ -39,13 +41,22 @@ export class DotEditLayoutComponent implements OnInit {
...
             ])
             .subscribe();
 
         this.route.data.pluck('pageView').subscribe((pageView: DotPageView) => {
             this.pageView = pageView;
-            this.initForm(pageView);
+
+            if (!this.isLayout()) {
+                this.showDialogSelection = true;
+            }
```
freddyucv: maybe we can move this for a method and give it a a good name

src/app/api/services/dot-messages-service.ts
fmontes: We don't need that because [arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) is a "special JS code.

src/app/portlets/dot-edit-page/layout/dot-edit-layout/dot-edit-layout.component.html
fmontes: Because a template

```anonymous: false```

Could be a layout and the only way to differentiate (in the front end) is by setting the name to null, BUT I just move it to the ts.

File: src/app/portlets/dot-edit-page/layout/dot-edit-layout/dot-edit-layout.component.ts [hunk 5]
```diff
@@ -39,13 +41,22 @@ export class DotEditLayoutComponent implements OnInit {
...
+                'editpage.layout.dialog.header'
             ])
             .subscribe();
 
         this.route.data.pluck('pageView').subscribe((pageView: DotPageView) => {
             this.pageView = pageView;
-            this.initForm(pageView);
+
+            if (!this.isLayout()) {
+                this.showTemplateLayoutSelectionDialog = true;
```
fmontes: I think with this new name and because we only do this here and once we don't need a method for this, I think it's pretty clear what this 3 lines does.

PR #507, opened 2018-01-09
Issue 13273 save layout and notifications

hmoreras:


src/app/view/components/_common/dot-global-message/dot-global-message.service.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    it('should set the default DotGlobalMessage Object for the Loading scenario', () => {
+        dotGlobalMessageService.loading();
+        listenerDotGlobalMessage.subscribe((event: DotEvent) => {
+            expect(event.data).toEqual({ value: 'Loading...', type: 'loading' });
+        });
+    });
+
+    it('', () => {
```
freddyucv: what do this test do?

src/app/view/components/_common/dot-global-message/dot-global-message.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Display text messages.
+     * @param {string} message
+     */
+    display(message?: string): void {
+        this.config = {
+            value: message ? message : this.dotMessageService.get('dot.common.message.loaded'),
+            life: 3000
```
freddyucv: this must be a class attribute

src/app/portlets/dot-edit-page/layout/dot-edit-layout/dot-edit-layout.component.ts
```diff
@@ -20,6 +24,9 @@ export class DotEditLayoutComponent implements OnInit {
     @ViewChild('templateName') templateName: ElementRef;
 
     form: FormGroup;
+    initialFormValue: FormGroup;
+    disableUpdateButton = true;
```
fmontes: Let's rename this property to ```canSave``` and let's use the same across the app... cc: @freddyucv @luisrojascr

src/app/portlets/dot-edit-page/layout/dot-edit-layout/dot-edit-layout.component.ts
```diff
@@ -125,11 +143,20 @@ export class DotEditLayoutComponent implements OnInit {
...
+        this.initForm();
+    }
+
     private initForm(): void {
         this.form = this.fb.group({
             title: this.isLayout() ? null : this.pageView.template.title,
             layout: this.fb.group({
-                body: this.pageView.layout.body || {},
+                body:
+                    this.dotEditLayoutService.getDotLayoutBody(
```
fmontes: Why is this change for?

src/app/view/components/_common/dot-global-message/dot-global-message.component.html
```diff
@@ -0,0 +1,4 @@
+<div [ngClass]="{'dot-global-message__visible': visibility}">
```
fmontes: Visible is a state, it's a modifier in BEM: http://getbem.com/naming/

src/app/view/components/_common/dot-global-message/dot-global-message.component.scss
```diff
@@ -0,0 +1,26 @@
+@import "primeng-theme/utils/theme-variables";
+
+:host{
+    display: flex;
+
+    div:first-child{
```
fmontes: You don't need to nest everything under host... that's why BEM class name are for.

Also the angular encapsulation takes care of this.

Fix the spacing.

You don't need ```div.xxx```

src/app/view/components/_common/dot-global-message/dot-global-message.component.scss
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    div.dot-global-message__visible{
+        opacity: 1;
+        transition:  $basic-speed * 2;
+    }
+
+    .dot-global-message__icon{
+        margin: 0 $basic-padding / 2;
+        font-size: 19px;
```
fmontes: There are variables for font sizes, check if there is one that fixes.

src/app/view/components/_common/dot-global-message/dot-global-message.component.scss
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    .dot-global-message__text{
+        font-size: $font-size-medium;
```
fmontes: This should be 13px, check zeplin

src/app/view/components/_common/dot-global-message/dot-global-message.component.scss
```diff
@@ -0,0 +1,26 @@
+@import "primeng-theme/utils/theme-variables";
+
+:host{
+    display: flex;
+
+    div:first-child{
+        opacity: 0;
+        transition:  $basic-speed * 2;
```
fmontes: You don't need to add the transition property in both states of element.

src/app/view/components/_common/dot-global-message/dot-global-message.component.scss
```diff
@@ -0,0 +1,26 @@
...
+
+    .dot-global-message__icon{
+        margin: 0 $basic-padding / 2;
+        font-size: 19px;
+    }
+
+    .dot-global-message__text{
+        font-size: $font-size-medium;
+        line-height: 19px;
+        vertical-align: top;
```
fmontes: You sure you need this?

src/app/view/components/_common/dot-global-message/dot-global-message.component.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+        dotEventsService.notify('dot-global-message', { value: 'test', type: 'loading', life: 3000 });
+        expect(component.message).toEqual({ value: 'test', type: 'fa fa-circle-o-notch fa-spin', life: 3000 });
+    });
+
+    it('should set visibility to false after 10 ms', () => {
+        dotEventsService.notify('dot-global-message', { value: 'test', life: 10 });
+        setTimeout(() => {
+            expect(component.visibility).toEqual(false);
+        }, 12);
+    });
```
fmontes: You need one to check a "sticky" message.

src/app/view/components/_common/dot-global-message/dot-global-message.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    message: DotGlobalMessage = { value: '' };
+
+    private icons = {
+        loading: 'fa fa-circle-o-notch fa-spin'
+    };
+
+    constructor(private dotEventsService: DotEventsService) {}
+
+    ngOnInit() {
+        this.dotEventsService.listen('dot-global-message').subscribe((event: DotEvent) => {
```
fmontes: You can do

```this.dotEventsService.listen('dot-global-message').filter(event => !!event.data)```

That way the subscribe callback will not be called unless there is something in the ```event.data```.

File: src/app/view/components/_common/dot-global-message/dot-global-message.service.ts (hunk 2 above)
Comment (fmontes) [reply src/app/view/components/_common/dot-global-message/dot-global-message.service.ts]: I think the opposite, we don't need ```config``` to be a class property, it doesn't represent anything in this class, we can just do:

```typescript
this.dotEventsService.notify('dot-global-message', {
    value: message ? message : this.dotMessageService.get('dot.common.message.loaded'),
    life: 3000
});
```

src/app/view/components/_common/dot-global-message/dot-global-message.service.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @export
+ * @class DotGlobalMessageService
+ */
+@Injectable()
+export class DotGlobalMessageService implements OnInit {
+    private config: DotGlobalMessage;
+
+    constructor(public dotMessageService: DotMessageService, private dotEventsService: DotEventsService) {}
+
+    ngOnInit(): void {
```
fmontes: Injectables classes doesn't have Lifecycle hooks, I don't think this is working correctly.

https://stackoverflow.com/questions/35110690/ngoninit-not-being-called-when-injectable-class-is-instantiated

src/app/view/components/dot-toolbar/dot-toolbar.component.scss
```diff
@@ -99,3 +99,7 @@
         }
     }
 }
+
+dot-global-message {
+    margin-right: 20px;
```
fmontes: use ```$basic-padding * 3```

src/app/view/components/dot-toolbar/dot-toolbar.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, ViewEncapsulation, Output, EventEmitter, Input } from '@angular/core';
+import { Component, ViewEncapsulation, Output, EventEmitter, Input, OnInit } from '@angular/core';
```
fmontes: We are not implementing OnInit in this class.

src/app/portlets/dot-edit-page/layout/dot-edit-layout-grid/dot-edit-layout-grid.component.ts
```diff
@@ -115,11 +115,11 @@ export class DotEditLayoutGridComponent implements OnInit, ControlValueAccessor
     }
 
     /**
-     * Event fired when the drag of a container ends, remove empty rows if any.
+     * Event fired when the drag or resize of a container ends, remove empty rows if any.
      *
      * @memberof DotEditLayoutGridComponent
      */
-    onDragStop(): void {
+    onDragOrReseizeStop(): void {
```
fmontes: Why just don't call this method like: ```updateModel``` because it's what it's doing it, right?

src/app/portlets/dot-edit-page/layout/dot-edit-layout/dot-edit-layout.component.ts
hmoreras: this gets the data that comes from the service and change the format to the result that will be send,  basically not wait the user to make changes but instead make it on load. and be able to compare if form is pristine

src/app/view/components/_common/dot-global-message/dot-global-message.component.scss
hmoreras: `$font-size-medium` is 13px.

src/app/view/components/_common/dot-global-message/dot-global-message.component.scss
hmoreras: yes otherwise don't do the vertical align correctly.

src/app/view/components/_common/dot-global-message/dot-global-message.component.spec.ts
hmoreras: I thought in that, but looks trivial if I always set the `visibility` to true.

src/app/view/components/_common/dot-global-message/dot-global-message.component.scss
hmoreras: there is none for 19px.

PR #534, opened 2018-01-30
dotCMS/core#13298 Implement dot-edit-page-toolbar functionality

fmontes:


src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+    }
+
+    /**
+     * Get the page HTML in live mode
+     *
+     * @param {string} url
+     * @returns {Observable<DotRenderedPage>}
+     * @memberof EditPageService
+     */
+    getLive(url: string): Observable<DotRenderedPage> {
```
fmontes: I don't know if the endpoint support this, I think is not.

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+        if (state.mode === PageMode.PREVIEW) {
+            pageMode = this.getPreview(page.pageUri);
+        } else if (state.mode === PageMode.EDIT) {
+            pageMode = this.getEdit(page.pageUri);
+        } else if (state.mode === PageMode.LIVE) {
+            pageMode = this.getPreview(page.pageUri);
+        }
+
+        if (lockUnlock && pageMode) {
+            return lockUnlock.mergeMap((lockState: string) => {
```
fmontes: Making sure I lock the page before getting it again.

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+     *
+     * @param {string} inode
+     * @returns {Observable<any>}
+     * @memberof PageViewService
+     */
+    lock(inode: string): Observable<any> {
+        return this.coreWebService.requestView({
+            method: RequestMethod.Put,
+            url: `content/lock/inode/${inode}`
+        });
```
freddyucv: I don't if we should return the ResponseView here, the end point return the follow:
{
      inode: inode,
      id:  id,
      message: "locked",
      return: 200
}

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+        if (state.lock === true) {
+            lockUnlock = this.lock(page.liveInode)
+                .pluck('bodyJsonObject')
+                .pluck('message');
+        } else if (state.lock === false) {
+            lockUnlock = this.unlock(page.liveInode)
+                .pluck('bodyJsonObject')
+                .pluck('message');
+
+        }
```
freddyucv: maybe we should create a private method for this

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+     * @returns {Observable<any>}
+     * @memberof EditPageService
+     */
+    setPageState(page: DotRenderedPage, state: DotEditPageState): Observable<DotRenderedPageState> {
+
+        const methods = [];
+        let lockUnlock;
+        let pageMode;
+
+        // When lock prop is not set is null
```
freddyucv: better something like: "When lock prop is not set then This should not be changed"
when we need comment to explain code then the code is not good enough, maybe is better use a enum here:

enum PageState {
    LOCK,
    UNLOCK
}

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+
+        }
+
+        if (state.mode === PageMode.PREVIEW) {
+            pageMode = this.getPreview(page.pageUri);
+        } else if (state.mode === PageMode.EDIT) {
+            pageMode = this.getEdit(page.pageUri);
+        } else if (state.mode === PageMode.LIVE) {
+            pageMode = this.getPreview(page.pageUri);
+        }
```
freddyucv: we should use this method: 'private get(url: string, mode: string)' here

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+     * @param {DotRenderedPage} page
+     * @param {DotEditPageState} state
+     * @returns {Observable<any>}
+     * @memberof EditPageService
+     */
+    setPageState(page: DotRenderedPage, state: DotEditPageState): Observable<DotRenderedPageState> {
+
+        const methods = [];
+        let lockUnlock;
+        let pageMode;
```
freddyucv: these are Observable

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+                    lockState: lockState
+                };
+            });
+        } else if (pageMode) {
+            return pageMode.map((dotRenderedPage: DotRenderedPage) => {
+                return {
+                    dotRenderedPage: dotRenderedPage
+                };
+            });
+        }
```
freddyucv: maybe we can clean up the code allways setting something to pageMode and lockUnLock, we can do something like:

if (state.lock === true) {
    ....
} else if (state.lock === false) {
   ....
} else {
    lockUnlock = Observable.of({})
}

and the same for pageMode

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+    }
+
+    private get(url: string, mode: string): Observable<DotRenderedPage> {
```
freddyucv: the second parameters must be PageMode not a string

src/app/api/services/page-view/page-view.service.ts
```diff
@@ -40,17 +40,4 @@ export class PageViewService {
             url: `v1/page/${pageIdentifier}/layout`
         }).pluck('entity');
     }
```
freddyucv: why we have 'PageViewService' and 'EditPageService'?

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs/Observable';
```
freddyucv: What happen with this class's test?

src/app/portlets/dot-edit-content/components/dot-edit-page-toolbar/dot-edit-page-toolbar.component.spec.ts
```diff
@@ -63,29 +63,43 @@ describe('DotEditPageToolbarComponent', () => {
...
+        component.page.pageUri = '/test/test';
         const pageUrlEl: HTMLElement = de.query(By.css('.edit-page-toolbar__page-url')).nativeElement;
         fixture.detectChanges();
 
         expect(pageUrlEl.textContent).toEqual('/test/test');
     });
 
-    it('should have a save button', () => {
+    it('should have a save button in edit mode', () => {
+        component.page.locked = true;
```
freddyucv: I think we need another test in edit mode and with the page unlock

src/app/portlets/dot-edit-content/components/dot-edit-page-toolbar/dot-edit-page-toolbar.component.spec.ts
```diff
@@ -138,6 +169,7 @@ describe('DotEditPageToolbarComponent', () => {
     });
 
     it('should disabled save button', () => {
+        component.page.locked = true;
         component.canSave = false;
```
freddyucv: why do canSave can turn into false?, I think we should tigger the event (o events) necessary for turn it into false and not just set it

src/app/portlets/dot-edit-content/components/dot-edit-page-toolbar/dot-edit-page-toolbar.component.spec.ts
```diff
@@ -159,89 +192,137 @@ describe('DotEditPageToolbarComponent', () => {
...
-    describe('moving through page states', () => {
+    it('should copy to clipboard url', () => {
+        spyOn(component, 'copyUrlToClipboard');
+        fixture.detectChanges();
+
+        const copyUrlButton: DebugElement = de.query(By.css('.edit-page-toolbar__copy-url'));
+
+        copyUrlButton.nativeElement.click();
+
+        expect(component.copyUrlToClipboard).toHaveBeenCalledTimes(1);
```
freddyucv: this method just check the the 'copyUrlToClipboard' method is call, but dont check f this is implement right
I think this test is not neccesary

src/app/portlets/dot-edit-content/components/dot-edit-page-toolbar/dot-edit-page-toolbar.component.ts
```diff
@@ -55,6 +54,39 @@ export class DotEditPageToolbarComponent implements OnInit {
...
+        txtArea.style.left = '0';
+        txtArea.style.opacity = '0';
+        txtArea.value = this.page.pageUri;
+        document.body.appendChild(txtArea);
+        txtArea.select();
+
+        try {
+            return document.execCommand('copy');
+        } catch (err) {
+            console.log('Oops, unable to copy');
```
freddyucv: shouldn't we show a error message in this case?

src/app/portlets/dot-edit-content/dot-edit-content.component.ts
```diff
@@ -137,58 +130,66 @@ export class DotEditContentComponent implements OnInit {
...
+            this.dotGlobalMessageService.display(this.dotMessageService.get('dot.common.message.saving'));
+        }
+
+        this.editPageService.setPageState(this.page, state).subscribe((dotRenderedPageState: DotRenderedPageState) => {
+            if (dotRenderedPageState.dotRenderedPage) {
+                this.setPage(dotRenderedPageState.dotRenderedPage);
+            }
+
+            if (state.lock !== null) {
+                this.dotGlobalMessageService.display(this.dotMessageService.get('dot.common.message.saved'));
```
freddyucv: We should show something like 'locked' insted

src/app/portlets/dot-edit-content/dot-edit-content.component.spec.ts
```diff
@@ -191,45 +195,116 @@ describe('DotEditContentComponent', () => {
...
-        spyOn(pageViewService, 'lock').and.callThrough();
+    it('should call the setPageState in the editPageService correctly', () => {
+        spyOn(editPageService, 'setPageState').and.callThrough();
+
+        fixture.detectChanges();
+
+        component.statePageHandler({
+            lock: false,
+            mode: null
+        });
```
freddyucv: I dont know if we should call the 'statePageHandler' method directly

src/app/portlets/dot-edit-page/shared/models/dot-rendered-page-state.model.ts
```diff
@@ -0,0 +1,6 @@
+import { DotRenderedPage } from './dot-rendered-page.model';
+
+export interface DotRenderedPageState {
+    dotRenderedPage: DotRenderedPage;
+    lockState: string;
```
freddyucv: This must be a enum

src/app/api/services/edit-page/edit-page.service.ts
fmontes: Tried to do it, it just too over engineering when the state it's just ```true/false``` so I rather to not send the prop and validate over that.

src/app/api/services/edit-page/edit-page.service.ts
fmontes: It returns ```bodyJsonObject``` so I'm going to pick from that.

src/app/api/services/edit-page/edit-page.service.ts
fmontes: Done

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
-    get(url: string): Observable<DotRenderedPage> {
+    /**
+     * Get the page HTML in edit mode
+     *
+     * @param {string} url
+     * @returns {Observable<DotRenderedPage>}
+     * @memberof EditPageService
+     */
+    getEdit(url: string): Observable<DotRenderedPage> {
+        return this.get(url, 'EDIT_MODE');
```
jdotcms: This should be an enum

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+
+    /**
+     * Get the page HTML in preview mode
+     *
+     * @param {string} url
+     * @returns {Observable<DotRenderedPage>}
+     * @memberof EditPageService
+     */
+    getPreview(url: string): Observable<DotRenderedPage> {
+        return this.get(url, 'PREVIEW_MODE');
```
jdotcms: enum

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,151 @@ import { Injectable } from '@angular/core';
...
+
+    /**
+     * Get the page HTML in live mode
+     *
+     * @param {string} url
+     * @returns {Observable<DotRenderedPage>}
+     * @memberof EditPageService
+     */
+    getLive(url: string): Observable<DotRenderedPage> {
+        return this.get(url, 'LIVE_MODE');
```
jdotcms: enum

src/app/api/services/edit-page/edit-page.service.ts
jdotcms: it is a legacy resource, it would good to start doing a v1 @wezell :)

src/app/api/services/edit-page/edit-page.service.ts
jdotcms: lockUnlock = (state.lock === true)?this.lock(page.liveInode).pluck('bodyJsonObject').pluck('message'):
                                                          this.unlock(page.liveInode).pluck('bodyJsonObject').pluck('message');

src/app/api/services/edit-page/edit-page.service.ts
jdotcms: This looks like state machine and
this could be handle by Map< PageMode, Method> map

map [PageMode.LIVE] = function (pageUri) {
      this.getEdit(pageUri)
}
....

then

pageMode = map[state.mode].(page.pageUri);

src/app/portlets/dot-edit-content/components/dot-edit-page-toolbar/dot-edit-page-toolbar.component.ts
jdotcms: shouldn't we have a Logger Service?

src/app/portlets/dot-edit-content/dot-edit-content.component.ts
jdotcms: +1 locked sounds more meningful

src/app/portlets/dot-edit-page/shared/models/dot-rendered-page-state.model.ts
jdotcms: +1

src/app/api/services/edit-page/edit-page.service.ts
fmontes: Done

src/app/api/services/edit-page/edit-page.service.ts
fmontes: Not exactly, because if ```state.lock``` is undefined we don't want to do any.

src/app/api/services/edit-page/edit-page.service.ts
fmontes: I like that @jdotcms forgot about that pattern, done.

src/app/portlets/dot-edit-content/components/dot-edit-page-toolbar/dot-edit-page-toolbar.component.ts
fmontes: We do have one (need to replace it because it's not AoT ready) but we shouldn't log this, we have a notification system for it.

src/app/portlets/dot-edit-content/dot-edit-content.component.ts
fmontes: Agreed, done.

src/app/api/services/edit-page/edit-page.service.ts
fmontes: Done

package.json
```diff
@@ -62,7 +62,7 @@
     "protractor": "~5.1.2",
     "ts-node": "~3.2.0",
     "tslint": "~5.3.2",
-    "typescript": "~2.3.3",
+    "typescript": "~2.4.2",
```
fmontes: Had to update it to use string in enums

src/app/api/services/edit-page/edit-page.service.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+        lastConnection = [];
+
+        this.injector = DOTTestBed.resolveAndCreate([
+            EditPageService
+        ]);
+
+        editPageService =  this.injector.get(EditPageService);
+        backend = this.injector.get(ConnectionBackend) as MockBackend;
+        backend.connections.subscribe((connection: any) => {
+            lastConnection.push(connection);
```
fmontes: with this method I can mock more than one connection

src/app/api/services/edit-page/edit-page.service.ts
fmontes: Done

src/app/api/services/edit-page/edit-page.service.ts
```diff
@@ -3,22 +3,159 @@ import { Injectable } from '@angular/core';
...
+        } else if (lock === false) {
+            return this.unlock(liveInode).pluck('message');
+        } else {
+            return null;
+        }
+    }
+
+    private getPageMode(mode: PageMode): (string) => Observable<DotRenderedPage> {
+        const map = {};
+        map[PageMode.PREVIEW] = (url: string) => this.getPreview(url);
```
fmontes: Instead of do a bunch of ```if```

src/app/portlets/dot-edit-content/components/dot-edit-page-toolbar/dot-edit-page-toolbar.component.ts
```diff
@@ -2,11 +2,14 @@ import { Component, OnInit, Input, EventEmitter, Output, ViewChild } from '@angu
...
+import { DotRenderedPage } from '../../../dot-edit-page/shared/models/dot-rendered-page.model';
+import { DotEditPageState } from '../../../../shared/models/dot-edit-page-state/dot-edit-page-state.model';
+import { DotGlobalMessageService } from '../../../../view/components/_common/dot-global-message/dot-global-message.service';
 
-export enum PageState {
-    EDIT,
-    PREVIEW,
-    LIVE
+export enum PageMode {
+    EDIT = 'EDIT_MODE',
```
fmontes: Only in tsc 2.4.X and up.

PR #540, opened 2018-02-05
Issue 13330 content types form restyling

luisrojascr:


src/app/portlets/content-types/form/content-types-form.component.spec.ts
```diff
@@ -86,9 +104,14 @@ describe('ContentTypesFormComponent', () => {
             el = de.nativeElement;
 
             dotcmsConfig = fixture.debugElement.injector.get(DotcmsConfig);
+            location = fixture.debugElement.injector.get(Location);
         })
     );
 
+    it('should open dialog by default', () => {
```
freddyucv: this test is not checking if the dialog is open

src/app/portlets/content-types/form/content-types-form.component.spec.ts
```diff
@@ -802,8 +732,4 @@ describe('ContentTypesFormComponent', () => {
             system: null
         });
     });
-
-    xit('should reset and collapse the form on ESC key', () => {
-        // TODO: waiting for Luis mock hotkeys.
-    });
```
freddyucv: I think we need a new test to check when the edit button is clicked, Also I think we need a test to check the submitForm method is called when the button is clicked

src/app/portlets/content-types/form/content-types-form.component.html
```diff
@@ -1,77 +1,108 @@
...
+        [closable]="false"> 
+    
+        <p-header class="container-fluid">
+            {{templateInfo.header}} {{i18nMessages['contenttypes.content.contenttype']}}
+        </p-header>
+    
+        <form [formGroup]="form" class="content-type__form container-fluid" (ngSubmit)="submitContent($event)" id="content-type-form" #formEl="ngForm" novalidate>
+            <div class="form__group">
+                <!-- TODO: Fix md-inputtext directive for floating labels -->
+                <!-- <span md-inputtext label="{{templateInfo.placeholder}}">
```
luisrojascr: I commented the md-inputtext directive implementation because it was not working and it was breaking the app.

src/app/api/services/dot-messages-service.ts
```diff
@@ -48,10 +48,10 @@ export class DotMessageService {
      * @returns {string}
      * @memberof DotMessageService
      */
-    get(key: string): string {
-        if (arguments && arguments.length > 1) {
+    get(key: string, ...args: string[]): string {
```
fmontes: This is the way to get "the rest" arguments of a method in TS

src/app/portlets/content-types/edit/content-types-edit.component.html
```diff
@@ -1,4 +1,4 @@
-<dot-content-type-layout [contentTypeId]="data?.id">
+<dot-content-type-layout [contentTypeId]="data?.id" *ngIf="isEditMode()">
```
fmontes: If is create mode I want to show nothing under the dialog.

src/app/portlets/content-types/edit/content-types-edit.component.html
```diff
@@ -7,48 +7,46 @@
...
-    <p-dialog
-        modal="modal"
-        width="500"
-        [dismissableMask]="true"
-        [(visible)]="show"
-        [responsive]="true"
-        [resizable]="false"
-        [draggable]="false"
-        [closable]="isEditMode()"
-        [height]="dialogHeight"
```
fmontes: Dialogs NEEDS to be responsive, we can't set the height as big as the window if the dialog content doesn't need to.

src/app/portlets/content-types/edit/content-types-edit.component.html
```diff
@@ -7,48 +7,46 @@
...
+
+    <dot-content-types-form
+        #form
+        [fields]="fields"
+        [data]="data"
+        (onSubmit)="handleFormSubmit($event)">
+    </dot-content-types-form>
+
+    <p-footer>
+        <button type="button" pButton label="Cancel" class="content-type__cancel cancel" (click)="cancelForm()"></button>
```
fmontes: ALWAYS add ```class="cancel"``` to all the cancel buttons in dialogs, it make it look black, we don't want them blue because is like a "neutral" action.

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
```diff
@@ -5,7 +5,7 @@ import { ContentTypesEditComponent } from './content-types-edit.component';
 import { CrudService } from '../../../api/services/crud/crud.service';
```
fmontes: This tests change a lot.

src/app/portlets/content-types/edit/content-types-edit.component.ts
```diff
@@ -68,7 +68,21 @@ export class ContentTypesEditComponent implements OnInit {
         ]).subscribe();
 
         this.setTemplateInfo();
-        this.dialogHeight = window.innerHeight - 50;
```
fmontes: Like I said: NEVER add a fix height to a dialog unless we know the content needs to scroll like in the iframe inside the dialog of edit page.

src/app/portlets/content-types/edit/content-types-edit.component.ts
```diff
@@ -77,33 +91,17 @@ export class ContentTypesEditComponent implements OnInit {
...
         this.dotMessageService.messageMap$.subscribe(() => {
-            const type = this.data.baseType.toLowerCase();
+            const type = this.contentTypesInfoService.getPrettyName(this.data.baseType);
 
             this.templateInfo = {
                 icon: this.contentTypesInfoService.getIcon(type),
                 header: this.isEditMode()
-                    ? this.dotMessageService.get('contenttypes.content.edit.contenttype')
-                    : this.dotMessageService.get('contenttypes.content.create.contenttype')
+                    ? this.dotMessageService.get('contenttypes.content.edit.contenttype', type)
```
fmontes: Remember we can do this.

src/app/portlets/content-types/form/content-types-form.component.html
```diff
@@ -1,61 +1,49 @@
...
+<form [formGroup]="form" class="content-type__form" id="content-type-form" novalidate>
     <div class="form__group">
-        <!-- TODO: Fix md-inputtext directive for floating labels -->
-        <span md-inputtext label="{{templateInfo.placeholder}}">
-            <input pInputText type="text" id="content-type-form-name" [style]="{'width':'100%'}" name="name" formControlName="name" tabindex="1" #name class="mousetrap" />
+        <span md-inputtext label="{{nameFieldLabel}}">
+            <input pInputText type="text" id="content-type-form-name" name="name" formControlName="name" tabindex="1" #name class="mousetrap" />
         </span>
-        <field-validation-message [message]="i18nMessages['contenttypes.form.name.error.required']" [field]="form.get('name')"></field-validation-message>
+        <field-validation-message [message]="dotMessageService.get('dot.common.message.field.required', nameFieldLabel)" [field]="form.get('name')"></field-validation-message>
```
fmontes: We can use the service directly.

src/app/portlets/content-types/form/content-types-form.component.spec.ts
```diff
@@ -1,14 +1,14 @@
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
```
fmontes: Same thing, tests here change a lot

src/app/portlets/content-types/form/content-types-form.component.ts
```diff
@@ -1,160 +1,93 @@
...
+ * @class ContentTypesFormComponent
+ * @implements {OnInit}
+ */
 @Component({
     providers: [SiteSelectorComponent],
     selector: 'dot-content-types-form',
     styleUrls: ['./content-types-form.component.scss'],
     templateUrl: 'content-types-form.component.html'
 })
-export class ContentTypesFormComponent extends BaseComponent implements OnInit, OnChanges, AfterViewInit, OnDestroy {
```
fmontes: I was able to get rid about all this, because the form now the form shows and hide in a dialog and I put an ```*ngIf``` in the dialog, which means that when the dialog hides the form is destroyed so I don't need to watch changes, everytime the dialog shows the form start on ```ngOnInit```.

src/app/portlets/content-types/edit/content-types-edit.component.html
```diff
@@ -7,48 +7,46 @@
...
+        <button type="button" pButton class="content-type__save" [label]="isEditMode() ? dotMessageService.get('contenttypes.action.update') : dotMessageService.get('contenttypes.action.create')" [disabled]="!form.canSave" (click)="form.submitForm()"></button>
```
fmontes: The disabled attr and the click is bind to the form properties and submit method... can't think a better way of doing this.

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
```diff
@@ -39,85 +46,167 @@ class TestContentTypeLayoutComponent {
...
+        spyOn(form, 'submitForm');
+        form.canSave = true;
+        fixture.detectChanges();
+
+        const saveButton = de.query(By.css('.content-type__save'));
+        saveButton.nativeElement.click();
+        expect(form.submitForm).toHaveBeenCalledTimes(1);
+    });
+
+    it('should bind save button disabled attribute to canSave property from the form', () => {
```
freddyucv: we dont should test it, we should test that when the form change the save button is enabled end disabled in otherwise

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
```diff
@@ -39,85 +46,167 @@ class TestContentTypeLayoutComponent {
...
+        fixture.detectChanges();
+
+        const saveButton = de.query(By.css('.content-type__save'));
+        expect(saveButton === null).toBe(false);
+        expect(saveButton.nativeElement.disabled).toBe(false);
+    });
+
+    it('should close the dialog and redirect', () => {
+        spyOn(dotRouterService, 'gotoPortlet');
+        spyOn(comp, 'cancelForm').and.callThrough();
```
freddyucv: we dont should test that this method is called, we should test that the dialog is closed

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
```diff
@@ -39,85 +46,167 @@ class TestContentTypeLayoutComponent {
...
-        const contentTypeForm = de.query(By.css('dot-content-type-fields-drop-zone'));
-        expect(contentTypeForm).toBeFalsy();
+    it('should have show form by default', () => {
+        const dialog = de.query(By.css('p-dialog'));
+        const contentTypeForm = de.query(By.css('dot-content-types-form'));
+
+        expect(comp.show).toBe(true);
+        expect(contentTypeForm === null).toBe(false);
+        expect(dialog === null).toBe(false);
     });
```
freddyucv: the last two test have the same description

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
```diff
@@ -39,85 +46,167 @@ class TestContentTypeLayoutComponent {
...
 
-    it('should NOT have Content Types Fields Drop Zone', () => {
-        const contentTypeForm = de.query(By.css('dot-content-type-fields-drop-zone'));
-        expect(contentTypeForm).toBeFalsy();
+    it('should have show form by default', () => {
+        const dialog = de.query(By.css('p-dialog'));
+        const contentTypeForm = de.query(By.css('dot-content-types-form'));
+
+        expect(comp.show).toBe(true);
```
freddyucv: we dont should set the show prop, we should trigger the events needed to make the dialog open

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
```diff
@@ -139,81 +228,139 @@ describe('ContentTypesEditComponent', () => {
...
+        clickEditButton();
+
+        spyOn(dotRouterService, 'gotoPortlet');
+
+        spyOn(comp, 'cancelForm').and.callThrough();
+        const cancelButton = de.query(By.css('.content-type__cancel'));
+        cancelButton.nativeElement.click();
+        fixture.detectChanges();
+
+        expect(comp.cancelForm).toHaveBeenCalledTimes(1);
```
freddyucv: I would remove this expect

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
```diff
@@ -39,85 +46,167 @@ class TestContentTypeLayoutComponent {
...
+
+    it('should have save button and disabled by default', () => {
+        const saveButton = de.query(By.css('.content-type__save'));
+        expect(saveButton === null).toBe(false);
+        expect(saveButton.nativeElement.disabled).toBe(true);
+    });
+
+    it('should submit form when save button is clicked', () => {
+        const form: ContentTypesFormComponent = de.query(By.css('dot-content-types-form')).componentInstance;
+        spyOn(form, 'submitForm');
```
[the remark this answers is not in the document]
fmontes: It is tested in the content-type-forms component.

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
fmontes: When the form change it change in the the content-type-form and that's the one in charge of set the canSave true or false and I use that property for the save button here in this component, I just need to make sure that the property is binded to the attr and I need to trust that the content-type-form is sending me the right value.

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
fmontes: I do validate that the diablog is close by doing this:

```expect(comp.show).toBe(false);```

Also there is other test where I validate that the dialog show or hide state is binded to the show property, like in this 3

https://github.com/dotCMS/core-web/blob/issue-13330-content-types-form-restyling/src/app/portlets/content-types/edit/content-types-edit.component.spec.ts#L129

https://github.com/dotCMS/core-web/blob/issue-13330-content-types-form-restyling/src/app/portlets/content-types/edit/content-types-edit.component.spec.ts#L185

https://github.com/dotCMS/core-web/blob/issue-13330-content-types-form-restyling/src/app/portlets/content-types/edit/content-types-edit.component.spec.ts#L192

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
fmontes: Im not setting that is the ngOnInit that set the show on true or false, in this line I'm just expecting.

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
fmontes: It's duplicated, fixing it.

src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
fmontes: Why?

PR #544, opened 2018-02-09
Issue 13541 prevent dd second contentlet

hmoreras:


src/app/portlets/dot-edit-content/services/html/iframe-edit-mode.js.ts
```diff
@@ -31,24 +31,38 @@ export const EDIT_PAGE_JS = `
...
+               el.dataset.dotBasetype === 'FORM' ||
+               target.dataset.dotAcceptTypes.indexOf(el.dataset.dotType) > -1;
+    }
+    
+    function checkIfMaxLimitNotReached(target){
+        return target.children.length < parseInt(target.dataset.maxContentlets, 10)
+    }
+    
+    function checkIfContentletIsUnique(el, target){    
+        return ! Array.from(target.querySelectorAll("div[data-dot-object='contentlet']"))
```
fmontes: Extra space in ```return ! Array```

src/app/portlets/dot-edit-content/services/html/iframe-edit-mode.js.ts
```diff
@@ -31,24 +31,38 @@ export const EDIT_PAGE_JS = `
...
+            .filter(node => {
```
fmontes: Does this work in IE11? I'm not sure if .filter is supported there.

The param "node" is a contentlet is better to use that for reading.

PR #575, opened 2018-03-15
dotCMS/core#13723 Content types UI issues (round 2)

fmontes:


src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -98,9 +98,6 @@ export class ContentTypesPortletComponent implements OnInit {
 
             this.actionHeaderOptions = {
                 primary: {
-                    command: ($event) => {
```
freddyucv: why do you remove it?

fmontes: That was reaaaally old from when we select the content type to create inside the creation page, now we have the submenu.

PR #576, opened 2018-03-15
Fix logic that handles switching sites when the current is archived

alfredo-dotcms:


src/app/view/components/_common/site-selector/site-selector.component.ts
```diff
@@ -85,6 +85,19 @@ export class SiteSelectorComponent implements OnInit, OnChanges {
      */
     handleSitesRefresh(): void {
         this.paginationService.getCurrentPage().subscribe((items) => {
+            const sites: any[] = items.splice(0);
```
alfredo-dotcms: since the interfase for SITE is outdated, the way to get filled the array with data is not specifying a type.

src/app/view/components/_common/site-selector/site-selector.component.ts
```diff
@@ -85,6 +85,19 @@ export class SiteSelectorComponent implements OnInit, OnChanges {
...
+            this.getSitesList();
+
+            // if current site not being listed on updated sites list (socket message) then switch to default site
+            if (!sites.find((site) => site.identifier === this.siteService.currentSite.identifier)) {
```
fmontes: This conditions is too complicated, when we find this kind of conditions we move it into a private method with a name that tells what is doing.

src/app/view/components/_common/site-selector/site-selector.component.ts
```diff
@@ -85,6 +85,19 @@ export class SiteSelectorComponent implements OnInit, OnChanges {
...
+                const defaultSite = sites.filter((site) => site.default === true);
```
fmontes: Same as before, you can do a private method like: ```getDefaultSite()```

src/app/view/components/_common/site-selector/site-selector.component.ts
alfredo-dotcms: done

src/app/view/components/_common/site-selector/site-selector.component.ts
alfredo-dotcms: done

src/app/view/components/_common/site-selector/site-selector.component.ts
```diff
@@ -85,6 +86,19 @@ export class SiteSelectorComponent implements OnInit, OnChanges {
...
         this.paginationService.getCurrentPage().subscribe((items) => {
+            this.sites = items.splice(0);
+            this.getSitesList();
+
+            // if current site not being listed on updated sites list (socket message) then switch to default site
+            if (!this.isCurrentSiteOnListOfAvailableSites(this.sites, this.siteService.currentSite)) {
+                const defaultSite = this.getDefaultSite(this.sites);
+                this.siteChange({
+                    hostname: defaultSite.hostname,
+                    type: undefined,
```
fmontes: Why is this for?

src/app/view/components/_common/site-selector/site-selector.component.ts
```diff
@@ -50,6 +50,7 @@ export class SiteSelectorComponent implements OnInit, OnChanges {
     currentSite: Observable<Site>;
     totalRecords: number;
     sitesCurrentPage: Site[];
+    sites: any[];
```
fmontes: Should this be ```Site[]````?

src/app/view/components/_common/site-selector/site-selector.component.ts
```diff
@@ -157,4 +169,14 @@ export class SiteSelectorComponent implements OnInit, OnChanges {
             });
         }
     }
+
+    private isCurrentSiteOnListOfAvailableSites(sites: any[], currentSite: Site): boolean {
+        return sites.find((site) => {
```
fmontes: You can use explicit return here

```
return sites.find((site) => site.identifier === currentSite.identifier && !site.archived)
```

src/app/view/components/_common/site-selector/site-selector.component.ts
```diff
@@ -157,4 +169,14 @@ export class SiteSelectorComponent implements OnInit, OnChanges {
...
+            return (site.identifier === currentSite.identifier && !site.archived);
+        });
+    }
+
+    private getDefaultSite(sites: any[]): Site {
+        return sites.filter((site) => site.default === true)[0];
```
fmontes: You don't need to compare ```=== true```, you can just do

```
return sites.filter((site) => site.default)[0];
```

src/app/view/components/_common/site-selector/site-selector.component.ts
alfredo-dotcms: if the current site has been archived, then we change the current site in the dropdown for the default site

src/app/view/components/_common/site-selector/site-selector.component.ts
alfredo-dotcms: the old interface for site only has 3 attributes, in order to use it with the new fields from the backend that interface needs to be updated with all the attributes which are around 20

src/app/view/components/_common/site-selector/site-selector.component.ts
alfredo-dotcms: done

src/app/view/components/_common/site-selector/site-selector.component.ts
alfredo-dotcms: done

src/app/view/components/_common/site-selector/site-selector.component.ts
freddyucv: why not just  do some like this: 'this.siteChange(defaultSite)'?

src/app/view/components/_common/site-selector/site-selector.component.ts
freddyucv: I think we can use Site anyway, the miss attributes just will be ignore

alfredo-dotcms: I did that test before and since the interface doesn't has all the attributes, when paginationService.getCurrentPage() returns the objects, those objects contains more attributes that don't match the interface and in the end, sites[] variable endup with length 0

src/app/view/components/_common/site-selector/site-selector.component.ts
alfredo-dotcms: I'll try this, but i'm not sure if what I mentioned on https://github.com/dotCMS/core-web/pull/576#discussion_r175796308 will happen...

src/app/view/components/_common/site-selector/site-selector.component.ts
```diff
@@ -85,6 +86,14 @@ export class SiteSelectorComponent implements OnInit, OnChanges {
      */
     handleSitesRefresh(): void {
         this.paginationService.getCurrentPage().subscribe((items) => {
+            this.sites = items.splice(0);
+            this.getSitesList();
+
+            // if current site not being listed on updated sites list (socket message) then switch to default site
```
freddyucv: I think we really need is a TODO instead, some thing like: 'look for a better implementation...' (we need a refactoring into site.service)

fmontes: This will not work if the list have more that one page.

src/app/view/components/dot-toolbar/dot-toolbar.component.ts
```diff
@@ -1,18 +1,44 @@
...
+     * @memberof SiteSelectorComponent
+     */
+    handleSitesRefresh(site: any): void {
+        // if current site is equal to the site updated (socket message) then switch to default site
+        if (site.hostname === this.siteService.currentSite.hostname && site.archived) {
+            // this.siteChange(this.getDefaultSite(this.sites));
+            this.siteService.getDefaultSite().subscribe((defaultSite: Site) => {
+                console.log('---defaultSite', defaultSite);
+                this.siteChange(defaultSite);
+                // this.currentSite = Observable.of(data);
```
freddyucv: remove all the commented code

src/app/view/components/dot-toolbar/dot-toolbar.component.ts
```diff
@@ -8,11 +8,32 @@ import { IframeOverlayService } from '../_common/iframe/service/iframe-overlay.s
...
     @Output() mainButtonClick: EventEmitter<MouseEvent> = new EventEmitter();
 
-    constructor(public iframeOverlayService: IframeOverlayService, private siteService: SiteService) {}
+    constructor(
+        public iframeOverlayService: IframeOverlayService,
+        private siteService: SiteService,
+    ) {}
+
+    ngOnInit(): void {
+        this.siteService.refreshSites$.subscribe((_site: Site) => this.handleSitesRefresh(_site));
```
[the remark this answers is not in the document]
freddyucv: really here we need just listen by ARCHIVE_SITE, maybe here we should use the dotcms_events-service directly

alfredo-dotcms:
Updated 'dotcms-js' package version

PR #582, opened 2018-03-19
Add command to update Workflows actions in a page asset

alfredo-dotcms:


src/app/api/services/dot-workflow/dot-workflow.service.ts
```diff
@@ -44,4 +44,20 @@ export class DotWorkflowService {
...
     }
+
+    /**
+     * Updates the workflow actions for a page asset
+     *
+     * @param {string} inode
+     * @returns {Observable<any[]>}
+     * @memberof DotWorkflowService
+     */
+    updateWorkflowActions(inode: string, actionId: string): Observable<any[]> {
```
fmontes: This is not "update" the action is fire. So maybe we can call this method: ```fireWorkflowAction```

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
```diff
@@ -13,20 +13,24 @@ export class DotEditPageWorkflowsActionsComponent implements OnInit {
...
     }
 
     private getWorkflowOptions(workflows: DotWorkflowAction[]): MenuItem[] {
         return workflows.map((workflow: DotWorkflowAction) => {
             return {
-                label: workflow.name
+                label: workflow.name,
+                command: () => {
+                    const request: Observable<any[]> = this.workflowsService.updateWorkflowActions(this.inode, workflow.id);
+                    request.subscribe();
```
fmontes: As far as I understand how Worflows actions work, after you trigger one, you need to update the ```workflowsMenuActions``` becuase they can change if you move to another step of the workflow by fire the actions.

Does this fire endpoint return something? maybe the new actions to fire? Maybe @jdotcms can give us more information here.

Also, what happen is there is an error in the request here? you need to catch that.

I created this service for that: https://github.com/dotCMS/core-web/blob/master/src/app/api/services/dot-http-error-manager/dot-http-error-manager.service.ts

Here is an use case: https://github.com/dotCMS/core-web/blob/ddc1f0f76b31fdacf26777463f665aa1f74b25ea/src/app/portlets/content-types/edit/content-types-edit-resolver.service.ts#L41

src/app/test/dot-workflow-service.mock.ts
```diff
@@ -2,7 +2,7 @@ import { Observable } from 'rxjs/Observable';
 import { DotWorkflowAction } from '../shared/models/dot-workflow-action/dot-workflow-action.model';
 
 export class DotWorkflowServiceMock {
-    getContentWorkflowActions(): Observable<DotWorkflowAction[]> {
+    getContentWorkflowActions(iNode: string): Observable<DotWorkflowAction[]> {
```
fmontes: inode is one word and is not Apple related 🤣

src/app/api/services/dot-workflow/dot-workflow.service.ts
freddyucv: why any[]?, I think this end point return a contentlet, We have to create a contentlet interface

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
freddyucv: I think the end point return a contentelet (I don't know why), so we should call this method again:

https://github.com/dotCMS/core-web/blob/827ffcdcf8dad3364eb2aef1976bd77efce471d6/src/app/api/services/dot-workflow/dot-workflow.service.ts#L39

Also we should print some message in the GlobalMessage

src/app/api/services/dot-workflow/dot-workflow.service.ts
freddyucv: What happen with this method's test?

src/app/api/services/dot-workflow/dot-workflow.service.spec.ts
```diff
@@ -97,4 +97,13 @@ describe('DotWorkflowService', () => {
         ]);
         expect(this.lastConnection.request.url).toContain(`v1/workflow/contentlet/${inode}/actions`);
     });
+
+    it('should fire workflows page action', () => {
+        const inode = 'cc2cdf9c-a20d-4862-9454-2a76c1132123';
+        const actionId = '44d4d4cd-c812-49db-adb1-1030be73e69a';
+        spyOn(this.dotWorkflowService, 'fireWorkflowAction');
+        this.dotWorkflowService.fireWorkflowAction(inode, actionId);
```
fmontes: This test is not testing right... you are calling the method and then check if it was called and that's not what we need to test, because if I go to the service and change the url the request will failed but the test won't, what you need to test is the endpoint was called correctly, check the url and the method, ex:

https://github.com/dotCMS/core-web/blob/issue-13774-login-as-iframes/src/app/portlets/content-types/fields/service/field.service.spec.ts#L121

https://github.com/dotCMS/core-web/blob/issue-13774-login-as-iframes/src/app/api/services/paginator/paginator.service.spec.ts#L101

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
```diff
@@ -13,20 +14,47 @@ export class DotEditPageWorkflowsActionsComponent implements OnInit {
...
+                command: () => {
+                    const currentMenuActions = this.workflowsMenuActions;
+                    this.workflowsMenuActions = this.workflowsService
+                        .fireWorkflowAction(this.inode, workflow.id)
+                        .pluck('inode')
+                        .catch((error) => Observable.of(null))
+                        .mergeMap((inode: string) => {
+                            this.inode = inode || this.inode;
+                            return this.workflowsService.getContentWorkflowActions(this.inode).map((newWorkflows: DotWorkflowAction[]) => {
+                                if (newWorkflows.length === 0) {
```
fmontes: Please add a TODO here  beacause this will probably change if the endpoint gets updated.

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.spec.ts
```diff
@@ -51,6 +63,35 @@ describe('DotEditPageWorkflowsActionsComponent', () => {
...
         fixture.detectChanges();
         const actionsButton: SplitButton = de.query(By.css('.edit-page-toolbar__actions')).componentInstance;
-        expect(actionsButton.model).toEqual([{ label: 'Assign Workflow' }, { label: 'Save' }, { label: 'Save / Publish' }]);
+
+        expect(actionsButton.model[0].label).toEqual('Assign Workflow');
+        expect(actionsButton.model[1].label).toEqual('Save');
+        expect(actionsButton.model[2].label).toEqual('Save / Publish');
     });
+
+    it('should trigger the methods in the action buttons', () => {
```
fmontes: AWESOME test! 👏

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
```diff
@@ -13,20 +14,47 @@ export class DotEditPageWorkflowsActionsComponent implements OnInit {
...
-                label: workflow.name
+                label: workflow.name,
+                command: () => {
+                    const currentMenuActions = this.workflowsMenuActions;
+                    this.workflowsMenuActions = this.workflowsService
+                        .fireWorkflowAction(this.inode, workflow.id)
+                        .pluck('inode')
+                        .catch((error) => Observable.of(null))
+                        .mergeMap((inode: string) => {
+                            this.inode = inode || this.inode;
```
freddyucv: if the fireWorkflowAction method throw a error then we just reload the before inode's workflow actions?

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
freddyucv: when the end point return 'newWorkflows.length === 0'? when the inode doesn't exists?

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.spec.ts
```diff
@@ -51,6 +63,35 @@ describe('DotEditPageWorkflowsActionsComponent', () => {
...
+        firstButton.click();
+        expect(dotWorkflowService.fireWorkflowAction).toHaveBeenCalledWith(component.inode, workflowsActions[0].id);
+
+        secondButton.click();
+        expect(dotWorkflowService.fireWorkflowAction).toHaveBeenCalledWith(component.inode, workflowsActions[1].id);
+
+        thirdButton.click();
+        expect(dotWorkflowService.fireWorkflowAction).toHaveBeenCalledWith(component.inode, workflowsActions[2].id);
+
+        expect(dotWorkflowService.fireWorkflowAction).toHaveBeenCalledTimes(3);
```
freddyucv: We have to check that the 'dotWorkflowService.getContentWorkflowActions' method have to be call a second times after the fireWorkflowAction is returned

src/app/test/dot-workflow-service.mock.ts
```diff
@@ -60,4 +60,8 @@ export class DotWorkflowServiceMock {
             }
         ]);
     }
+
+    fireWorkflowAction(): Observable<any[]> {
+        return Observable.of([]);
```
freddyucv: we need return something here to test this case:

https://github.com/dotCMS/core-web/pull/582/files#diff-6a04a173805dcfecbbb2ca4256501928R34

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
fmontes: Because the endpoint returns a 200 with an empty array, I already told the Jons that in that case should return a 404

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
fmontes: Yes

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
freddyucv: but the empty array means that the inode doesn't have any actions or that the inode doesn't exists

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
freddyucv: I think we should add a TODO here, this is not a good solution in the case the fire request catch

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.spec.ts
```diff
@@ -51,6 +63,35 @@ describe('DotEditPageWorkflowsActionsComponent', () => {
     it('should set action split buttons params', () => {
         fixture.detectChanges();
         const actionsButton: SplitButton = de.query(By.css('.edit-page-toolbar__actions')).componentInstance;
-        expect(actionsButton.model).toEqual([{ label: 'Assign Workflow' }, { label: 'Save' }, { label: 'Save / Publish' }]);
+
+        expect(actionsButton.model[0].label).toEqual('Assign Workflow');
```
hmoreras: make a single expect a compare with the array of strings. !?

src/app/api/services/dot-workflow/dot-workflow.service.ts
```diff
@@ -44,4 +44,20 @@ export class DotWorkflowService {
...
+
+    /**
+     * Updates the workflow actions for a page asset
+     *
+     * @param {string} inode
+     * @returns {Observable<any>} // contentlet
+     * @memberof DotWorkflowService
+     */
+    fireWorkflowAction(inode: string, actionId: string): Observable<any> {
+        return this.coreWebService
```
jdotcms: I would be nice to have something such as coreWebService.put (url, body),  post(url, boy), delete(url), get(url)

src/app/test/dot-workflow-service.mock.ts
alfredo-dotcms: this method returns a contentlet, the test for the method itself of the dot-workflow-service.ts file is in dot-workflow.service.spec.ts

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.ts
fmontes: Yes, but if we send an inode that doesn't exist still return 200 with []

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.component.html
```diff
@@ -2,5 +2,6 @@
     flat 
     class="edit-page-toolbar__actions"
     [label]="label"
-    [model]="workflowsMenuActions | async">
+    [model]="workflowsMenuActions | async"
+    [disabled]="!(workflowsMenuActions | async)?.length">
```
fmontes: If no actions, disabled button.

src/app/portlets/dot-edit-page/content/components/dot-edit-page-toolbar/dot-edit-page-toolbar.component.spec.ts
```diff
@@ -92,6 +94,10 @@ describe('DotEditPageToolbarComponent', () => {
                         provide: DotWorkflowService,
                         useClass: DotWorkflowServiceMock
                     },
+                    {
+                        provide: LoginService,
```
fmontes: All toolbar tests were broken now, because the action components needs more providers.

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.spec.ts
```diff
@@ -44,59 +46,80 @@ describe('DotEditPageWorkflowsActionsComponent', () => {
...
         fixture = testbed.createComponent(DotEditPageWorkflowsActionsComponent);
+        de = fixture.debugElement;
+
         component = fixture.componentInstance;
         component.label = 'ACTIONS';
         component.inode = 'cc2cdf9c-a20d-4862-9454-2a76c1132123';
-        de = fixture.debugElement;
-    });
 
-    it('should have a workflow actions element', () => {
```
fmontes: Strying DRY by moving things in the test to the beforeEach

src/app/test/dot-test-bed.ts
```diff
@@ -42,7 +43,10 @@ export class DOTTestBed {
             DotDialogService,
             DotEventsService,
             DotGlobalMessageService,
+            DotHttpErrorManagerService,
```
fmontes: Add this to the testbed because is global.

src/app/test/dot-workflow-service.mock.ts
```diff
@@ -1,64 +1,66 @@
 import { Observable } from 'rxjs/Observable';
 import { DotWorkflowAction } from '../shared/models/dot-workflow-action/dot-workflow-action.model';
 
+export const mockWorkflows = [
```
fmontes: Create this mock so we can use it in the tests.

src/app/api/services/dot-workflow/dot-workflow.service.spec.ts
```diff
@@ -97,4 +97,17 @@ describe('DotWorkflowService', () => {
         ]);
         expect(this.lastConnection.request.url).toContain(`v1/workflow/contentlet/${inode}/actions`);
     });
+
+    it('should fire workflows page action', () => {
+        const inode = 'cc2cdf9c-a20d-4862-9454-2a76c1132123';
+        const actionId = '44d4d4cd-c812-49db-adb1-1030be73e69a';
+        spyOn(this.dotWorkflowService, 'fireWorkflowAction').and.callThrough();
+        const test = this.dotWorkflowService.fireWorkflowAction(inode, actionId).subscribe((res) => {
```
hmoreras: there is no need of the const test, right ?

src/app/api/services/dot-workflow/dot-workflow.service.spec.ts
```diff
@@ -3,7 +3,7 @@ import { DotWorkflowService } from './dot-workflow.service';
 import { MockBackend } from '@angular/http/testing';
 import { ConnectionBackend, ResponseOptions, Response } from '@angular/http';
 
-describe('DotWorkflowService', () => {
+fdescribe('DotWorkflowService', () => {
```
fmontes: fdescribe

src/app/portlets/dot-edit-page/content/components/dot-edit-page-workflows-actions/dot-edit-page-workflows-actions.spec.ts
```diff
@@ -1,19 +1,30 @@
...
 import { DotWorkflowService } from '../../../../../api/services/dot-workflow/dot-workflow.service';
+import { DotRouterService } from '../../../../../api/services/dot-router/dot-router.service';
+import { DotHttpErrorManagerService } from '../../../../../api/services/dot-http-error-manager/dot-http-error-manager.service';
 import { DotEditPageWorkflowsActionsComponent } from './dot-edit-page-workflows-actions.component';
-import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
-import { RouterTestingModule } from '@angular/router/testing';
+import { mockResponseView } from '../../../../../test/response-view.mock';
 
-describe('DotEditPageWorkflowsActionsComponent', () => {
+fdescribe('DotEditPageWorkflowsActionsComponent', () => {
```
fmontes: fdescribe

PR #583, opened 2018-03-19
fixing test

freddyucv:


src/app/test/dot-test-bed.ts
```diff
@@ -49,7 +50,8 @@ export class DOTTestBed {
             LoggerService,
             SocketFactory,
             StringUtils,
-            UserModel
+            UserModel,
+            DotRouterService
```
fmontes: Now we have to remove all DotRouterService from all the other tests.

PR #608, opened 2018-04-12
issue-13977-allow-multiple-container

hmoreras:


src/app/view/components/dot-container-selector/dot-container-selector.component.html
```diff
@@ -10,7 +10,8 @@
             [totalRecords]="paginationService.totalRecords"
             [labelPropertyName]="['name', 'parentPermissionable.hostname']"
             placeholder="{{ dotMessageService.get('editpage.container.add.label') }}"
-            persistentPlaceholder="true">
+            persistentPlaceholder="true"
+            [multiple]="multiple">
```
freddyucv: the value here is true

hmoreras: it will depend, of what the user send.. in this case we are sending true in `dot-edit-layout-grid.component.html`

PR #622, opened 2018-04-17
Issuse 13982 enter to submit form

hmoreras:


src/app/portlets/content-types/form/content-types-form.component.spec.ts
```diff
@@ -493,44 +494,64 @@ describe('ContentTypesFormComponent', () => {
...
+                folder: null,
+                system: null,
+                workflow: ['85c1515c-c4f3-463c-bac2-860b8fcacc34']
+            });
+        });
+        it('should submit form correctly on Enter', () => {
+            const form = fixture.debugElement.query(By.css('form'));
+            form.nativeElement.dispatchEvent(new KeyboardEvent('keyup', { key: 'Enter' }));
+            expect(comp.submitForm).toHaveBeenCalledTimes(1);
+            expect(data).toEqual({
```
fmontes: I think this expect is not neccessary here because this is being tested in the previous test, but is ok.

src/app/portlets/content-types/form/content-types-form.component.ts
```diff
@@ -132,7 +132,7 @@ export class ContentTypesFormComponent implements OnInit {
      * @memberof ContentTypesFormComponent
      */
     submitForm(): void {
-        if (this.form.valid) {
+        if (this.form.valid && this.canSave) {
```
freddyucv: We just need this.canSave here because this check this.form.valid

https://github.com/dotCMS/core-web/blob/54399744abcf74ae95330da9eba988f255043281/src/app/portlets/content-types/form/content-types-form.component.ts#L150

freddyucv: don't we need a test to check this change?

PR #642, opened 2018-05-02
Issue 14208 site change redirect to site browser

alfredo-dotcms:


src/app/view/components/dot-toolbar/dot-toolbar.component.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+        de = fixture.debugElement;
+        el = de.nativeElement;
+        dotRouterService = fixture.debugElement.injector.get(DotRouterService);
+        siteService = fixture.debugElement.injector.get(SiteService);
+    }));
+
+    it('should call to "goToSiteBrowser" in "DotRouterService" when the "siteChange" method is actioned', () => {
+        spyOn(dotRouterService, 'goToSiteBrowser');
+
+        comp.siteChange(siteMock);
```
fmontes: We need to test the binding of the ```dot-site-selector``` because as this tests is right now... if I remove the ```(change)="siteChange($event)"``` from the html of this component the test will not break.

src/app/view/components/dot-toolbar/dot-toolbar.component.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        el = de.nativeElement;
+        dotRouterService = fixture.debugElement.injector.get(DotRouterService);
+        siteService = fixture.debugElement.injector.get(SiteService);
+    }));
+
+    it('should trigger "siteChange" method when actioned', () => {
+        const siteSelector: DebugElement = fixture.debugElement.query(By.css('dot-site-selector'));
+        spyOn(comp, 'siteChange').and.callThrough();
+        siteSelector.triggerEventHandler('change', { value: siteMock });
+        expect(comp.siteChange).toHaveBeenCalledWith({ value: siteMock });
```
freddyucv: We don't need check this

src/app/view/components/dot-toolbar/dot-toolbar.component.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    });
+
+    it('should call to "goToSiteBrowser" in "DotRouterService" when the "siteChange" method is actioned', () => {
+        spyOn(dotRouterService, 'goToSiteBrowser');
+        comp.siteChange(siteMock);
+        expect(dotRouterService.goToSiteBrowser).toHaveBeenCalled();
+    });
```
freddyucv: this two test should be only one

PR #651, opened 2018-05-08
dotCMS/core#14165 Esc key should back to listing from content type create

fmontes:


src/app/portlets/content-types/edit/content-types-edit.component.spec.ts
```diff
@@ -156,6 +178,15 @@ describe('ContentTypesEditComponent create mode', () => {
...
+
+        testHotKeysMock.callback(['esc']);
+        fixture.detectChanges();
+
+        expect(comp.cancelForm).toHaveBeenCalledTimes(1);
+    });
+
     it('should NOT have dot-content-type-layout', () => {
         const contentTypeLayout = de.query(By.css('dot-content-type-layout'));
         expect(contentTypeLayout === null).toBe(true);
```
freddyucv: I think we need a test to check that the 'cancelForm' method is not called when editMode is false (if it doesn't exists yet)

fmontes: In fact... it have to be called because the cancelForm close the dialog but It might redirect or not... and that's being tested already here: https://github.com/dotCMS/core-web/blob/b69b6da32a1e77c350c5c7a46f276656e11a0e17/src/app/portlets/content-types/edit/content-types-edit.component.spec.ts#L383

PR #653, opened 2018-05-08
Issue 14129 login as admin user not handling wrong password

alfredo-dotcms:


src/app/view/components/login-as/login-as.ts
```diff
@@ -17,10 +17,12 @@ import { DotNavigationService } from '../dot-navigation/dot-navigation.service';
 export class LoginAsComponent extends BaseComponent implements OnInit {
     @Output() cancel = new EventEmitter<boolean>();
     @Input() visible: boolean;
+    @ViewChild('password') el: ElementRef;
```
fmontes: ```passwordEl```, ```el``` is too generic.

PR #657, opened 2018-05-14
Issue 14310 edit page nav component needs to update if the page have layout or not

alfredo-dotcms:


src/app/portlets/dot-edit-page/main/dot-edit-page-main/dot-edit-page-main.component.ts
```diff
@@ -11,9 +11,20 @@ import { DotRenderedPageState } from '../../shared/models/dot-rendered-page-stat
 export class DotEditPageMainComponent implements OnInit {
     pageState: Observable<DotRenderedPageState>;
 
-    constructor(private route: ActivatedRoute) {}
+    constructor(private route: ActivatedRoute, private router: Router) {
+        router.events.subscribe((val) => {
```
fmontes: Why not just:

```this.pageState = router.events.pipe(filter(...), mergeMap(this.route.data.pluck('content')))```

?

File: src/app/portlets/dot-edit-page/main/dot-edit-page-nav/dot-edit-page-nav.component.spec.ts [hunk 2]
```diff
@@ -93,19 +104,38 @@ describe('DotEditPageNavComponent', () => {
             const activeItem = fixture.debugElement.query(By.css('.edit-page-nav__item--active'));
             expect(activeItem.nativeElement.innerText).toContain('Content');
         });
+
+        it('should update menu items when new PageState', () => {
```
fmontes: This test and ```'should have basic menu items'``` should be in one describe, looks like we are repeating code thats should be in a ```beforeEach```

src/app/portlets/dot-edit-page/main/dot-edit-page-nav/dot-edit-page-nav.component.ts
```diff
@@ -38,6 +43,7 @@ export class DotEditPageNavComponent implements OnInit {
                 return this.dotLicenseService.isEnterpriseLicense();
             })
             .mergeMap((isEnterpriseLicense: boolean) => {
+                this.isEnterpriseLicense = isEnterpriseLicense;
                 return Observable.of(this.getNavItems(this.pageState, isEnterpriseLicense));
```
fmontes: I just notice this but, you don't need the ```Observable.of``` if instead of ```mergeMap``` you use ```map```.

PD. Use pipes.

src/app/portlets/dot-edit-page/main/dot-edit-page-nav/dot-edit-page-nav.component.ts
```diff
@@ -20,14 +20,19 @@ interface DotEditPageNavItem {
...
     @Input() pageState: DotRenderedPageState;
     model: Observable<DotEditPageNavItem[]>;
+    isEnterpriseLicense: boolean;
 
     constructor(private dotLicenseService: DotLicenseService, public dotMessageService: DotMessageService, public route: ActivatedRoute) {}
 
-    ngOnInit(): void {
-        this.model = this.dotMessageService
+    ngOnChanges(): void {
+        this.model = !this.model ? this.loadData() : Observable.of(this.getNavItems(this.pageState, this.isEnterpriseLicense));
```
fmontes: This looks problematic to me... I like the idea crate a class property of ```isEnterpriseLicense``` so we don't have a to do a request to get the license every time, but this could have some side effects, if the license is apply after this components loads the first time we'll never see the update. So I think we should call ```this.loadData()``` all the time, the messages are cached so that request will not happen (or at least it shouldn't)

fmontes: Also do you double check that this ```ngOnChanges``` is not getting called when we don't want to?

src/app/portlets/dot-edit-page/main/dot-edit-page-main/dot-edit-page-main.component.ts
```diff
@@ -1,19 +1,35 @@
...
+    private routeEventsSub: Subscription;
 
-    constructor(private route: ActivatedRoute) {}
+    constructor(private route: ActivatedRoute, private router: Router) {}
 
-    ngOnInit() {
+    ngOnInit(): void {
+        this.loadData();
+        this.routeEventsSub = this.router.events
+            .pipe(filter((anyEvent: any) => event instanceof NavigationEnd && event.url.indexOf('/edit-page/content') > -1))
```
freddyucv: I dont undestand what exactly is happend here:

"event instanceof NavigationEnd && event.url.indexOf('/edit-page/content')"

this code is not self explain so I think is better extract it into a method

src/app/portlets/dot-edit-page/main/dot-edit-page-main/dot-edit-page-main.component.ts
```diff
@@ -1,19 +1,35 @@
...
+    }
+
+    ngOnDestroy(): void {
+        this.routeEventsSub.unsubscribe();
+    }
+
+    private loadData(): void {
         this.pageState = this.route.data.pluck('content');
     }
 }
```
freddyucv: don't we need update any test to check this changes?

src/app/portlets/dot-edit-page/main/dot-edit-page-nav/dot-edit-page-nav.component.ts
```diff
@@ -20,26 +22,37 @@ interface DotEditPageNavItem {
...
 
-    ngOnInit(): void {
-        this.model = this.dotMessageService
+    ngOnChanges(changes: SimpleChanges): void {
+        if (changes.pageState.currentValue !== changes.pageState.previousValue) {
+            this.model = !this.model ? this.loadData() : Observable.of(this.getNavItems(this.pageState, this.isEnterpriseLicense));
+        }
+    }
+
+    private loadData(): Observable<DotEditPageNavItem[]> {
```
freddyucv: 'loadData' don't seem a good name for this method

src/app/portlets/dot-edit-page/main/dot-edit-page-nav/dot-edit-page-nav.component.ts
```diff
@@ -21,27 +23,39 @@ interface DotEditPageNavItem {
...
     @Input() pageState: DotRenderedPageState;
     model: Observable<DotEditPageNavItem[]>;
+    isEnterpriseLicense: boolean;
 
     constructor(private dotLicenseService: DotLicenseService, public dotMessageService: DotMessageService, public route: ActivatedRoute) {}
 
-    ngOnInit(): void {
-        this.model = this.dotMessageService
+    ngOnChanges(changes: SimpleChanges): void {
+        if (changes.pageState.currentValue !== changes.pageState.previousValue) {
```
fmontes: You sure you need to do this change? because as far as I understand if the ```ngOnChanges``` trigger is because a property change, right?

src/app/portlets/dot-edit-page/main/dot-edit-page-nav/dot-edit-page-nav.component.ts
```diff
@@ -31,13 +32,18 @@ export class DotEditPageNavComponent implements OnChanges {
...
 
     ngOnChanges(changes: SimpleChanges): void {
-        if (changes.pageState.currentValue !== changes.pageState.previousValue) {
+        if (this.layoutChanged(changes)) {
             this.model = !this.model ? this.loadNavItems() : Observable.of(this.getNavItems(this.pageState, this.isEnterpriseLicense));
         }
     }
 
-    private loadNavItems(): Observable<DotEditPageNavItem[]> {
+    private layoutChanged(changes: SimpleChanges): boolean {
```
freddyucv: this method signature should be 

 private layoutChanged(oldPageState: DotRenderedPageState, newPageState: DotRenderedPageState): boolean {

alfredo-dotcms: mmm, why?

alfredo-dotcms: inside the method it's used `changes.pageState.firstChange`, so according to your comment we should also add that third parameter :-S

src/app/portlets/dot-edit-page/main/dot-edit-page-nav/dot-edit-page-nav.component.ts
```diff
@@ -38,9 +38,9 @@ export class DotEditPageNavComponent implements OnChanges {
     }
 
     private layoutChanged(changes: SimpleChanges): boolean {
-        return changes.pageState.currentValue && changes.pageState.previousValue
-            ? !_.isEqual(changes.pageState.currentValue.layout, changes.pageState.previousValue.layout)
-            : true;
+        return changes.pageState.firstChange
```
fmontes: You can do

```typescript
return changes.pageState.firstChange || ...
```

But also, do we need ```_.isEqual```? since we know that one of those will be ```undefined``` all the time.

alfredo-dotcms: I think we do, because if you stay on the same page and it has the property `layout` & the user has enterprise license, and then you click on the button of layout (right side nav bar) then that condition will evaluate and you need to compare current and previous states to get a `false` value and don't do an unnecesary change

PR #662, opened 2018-05-18
Issue 14305 creating new content types the system workflow should be selected by default

alfredo-dotcms:


src/app/view/components/_common/dot-workflows-selector-field/dot-workflows-selector-field.component.spec.ts
```diff
@@ -147,11 +152,6 @@ describe('DotWorkflowsSelectorFieldComponent', () => {
             })
         );
 
-        it('should get value', () => {
-            fixtureHost.detectChanges();
-            expect(component.value).toEqual(['567', '890']);
-        });
-
```
freddyucv: why this test is remove?

src/app/view/components/_common/dot-workflows-selector-field/dot-workflows-selector-field.component.ts
```diff
@@ -42,14 +43,15 @@ export class DotWorkflowsSelectorFieldComponent implements ControlValueAccessor,
 
     ngOnInit() {
         this.options = this.dotMessageService.getMessages(['dot.common.select.workflows', 'dot.common.archived']).mergeMap(() => {
-            return this.dotWorkflowService
-                .get()
-                .do((workflows: DotWorkflow[]) => {
+            return this.dotWorkflowService.get().pipe(
```
fmontes: We can't do this here, because this way we're are setting ALL the instaces (even the ones that receive a ```formControl```) of ```DotWorkflowsSelectorFieldComponent``` with this default value and we don't want that, we want it only for when you are creating a content types, we need to do that [here](https://github.com/dotCMS/core-web/blob/master/src/app/portlets/content-types/form/content-types-form.component.ts#L82:L84).

We need to rename the method in the service to be like ```this.dotWorkflowService.getSystem``` update the implementation and set the value as we do right now.

PR #664, opened 2018-05-18
dotCMS/core#14327 Fix the device and personas dropdown default label

fmontes:


src/app/view/components/dot-device-selector/dot-device-selector.component.ts
```diff
@@ -21,8 +21,14 @@ export class DotDeviceSelectorComponent implements OnInit {
...
         this.options = this.dotMessageService
             .getMessages(['editpage.viewas.default.device'])
-            .mergeMap((messages: string[]) =>
-                this.dotDevicesService.get().pipe(map((devices: DotDevice[]) => this.setOptions(messages, devices)))
+            .mergeMap(() =>
+                this.dotDevicesService
+                    .get()
+                    .pipe(
+                        map((devices: DotDevice[]) =>
+                            this.setOptions(this.dotMessageService.get('editpage.viewas.default.device'), devices)
```
fmontes: Sooooooooo... we shouldn't been doing this

```this.dotMessageService.get('editpage.viewas.default.device')```

here, because in theory the ```.getMessages(['editpage.viewas.default.device'])``` should return that, but somehow is not working in FF and I don't have time right now to refactor the message service.

PR #665, opened 2018-05-18
dotCMS/core#14326 Bring back page model global function and prevent custom event breaks

fmontes:


src/app/portlets/dot-edit-page/content/services/html/iframe-edit-mode.js.ts
```diff
@@ -90,6 +90,8 @@ export const EDIT_PAGE_JS = `
             });
         }
     })
+
+    window.getDotNgModel = getDotNgModel;
```
fmontes: I remove it but we need it [here](https://github.com/dotCMS/core-web/blob/master/src/app/portlets/dot-edit-page/content/services/dot-edit-content-html/dot-edit-content-html.service.ts#L187)

src/app/view/components/_common/iframe/iframe-porlet-legacy/services/iframe-events-handler.service.ts
```diff
@@ -27,8 +27,16 @@ export class DotIframeEventsHandler {
...
-    handle(event: CustomEvent) {
-        this.handlers[event.detail.name](event);
+    /**
+     * Handle custom events from the iframe portlets
+     *
+     * @param {CustomEvent} event
+     * @memberof DotIframeEventsHandler
+     */
+    handle(event: CustomEvent): void {
+        if (this.handlers[event.detail.name]) {
```
fmontes: Just making sure non-handled events crash the app.

src/app/portlets/dot-edit-page/content/services/html/iframe-edit-mode.js.ts
fmontes: Also I tried to move this to the angular side, but we use it [here](https://github.com/dotCMS/core-web/blob/master/src/app/portlets/dot-edit-page/content/services/html/iframe-edit-mode.js.ts#L74) to send the model down with the event.

PR #672, opened 2018-05-24
dotCMS/core#13880 Content types: [Violation] 'click' handler took XXXms and auto focus add to bundle dropdown

fmontes:


fmontes:
I will create a custom menu component that is css positioned and show on click.

PR #680, opened 2018-06-01
Filter device selector

alfredo-dotcms:


src/app/portlets/dot-edit-page/content/dot-edit-content.component.spec.ts
```diff
@@ -788,7 +788,7 @@ describe('DotEditContentComponent', () => {
         });
     });
 
-    describe('Auto save', () => {
+    fdescribe('Auto save', () => {
```
fmontes: `f` here.

src/app/portlets/dot-edit-page/content/components/dot-edit-content-view-as-toolbar/dot-edit-content-view-as-toolbar.component.spec.ts
```diff
@@ -227,7 +227,7 @@ describe('DotEditContentViewAsToolbarComponent', () => {
             expect(whatsChanged).toBe(null);
         });
 
-        describe('events', () => {
+        fdescribe('events', () => {
```
fmontes: `f` here.

src/app/view/components/dot-device-selector/dot-device-selector.component.ts
```diff
@@ -21,14 +21,19 @@ export class DotDeviceSelectorComponent implements OnInit {
...
-                    .get()
-                    .pipe(
-                        map((devices: DotDevice[]) =>
-                            this.setOptions(this.dotMessageService.get('editpage.viewas.default.device'), devices)
+            .pipe(
+                mergeMap(() =>
+                    this.dotDevicesService
+                        .get()
+                        .pipe(
+                            flatMap((devices: DotDevice[]) => devices),
```
freddyucv: Are you sure that we need a flatMap here?

src/app/view/components/dot-device-selector/dot-device-selector.component.ts
```diff
@@ -21,14 +21,19 @@ export class DotDeviceSelectorComponent implements OnInit {
...
+                            filter((device: DotDevice) => +device.cssHeight > 0 && +device.cssWidth > 0),
+                            toArray(),
+                            map((devices: DotDevice[]) =>
+                                this.setOptions(this.dotMessageService.get('editpage.viewas.default.device'), devices)
+                            )
```
freddyucv: I think it shoud be a tap

src/app/test/dot-device.mock.ts
```diff
@@ -1,8 +1,16 @@
...
+        name: 'iphone',
+        inode: '1'
+    },
+    {
+        cssHeight: '0',
+        cssWidth: '0',
+        name: 'bad device',
+        inode: '2'
+    }
+];
```
freddyucv: shouldn't we have one with cssHeight and/or cssWidth equals to null?

alfredo-dotcms: If it's null, empty string or a zero (string or number) if will be casted to 0 (number). So not sure if needed!

src/app/view/components/dot-device-selector/dot-device-selector.component.ts
alfredo-dotcms: We need to convert the data so the filter operator can work with the data correctly

src/app/test/dot-device.mock.ts
fmontes: What is coming from the server when those fields are empty?

alfredo-dotcms: empty string

src/app/view/components/dot-device-selector/dot-device-selector.component.ts
alfredo-dotcms: no, we need to return the options

src/app/test/dot-device.mock.ts
fmontes: Ok, I think we should mock that.

PR #681, opened 2018-06-01
Fix unit test

alfredo-dotcms:


src/app/portlets/dot-edit-page/content/components/dot-edit-content-view-as-toolbar/dot-edit-content-view-as-toolbar.component.spec.ts
```diff
@@ -222,6 +222,7 @@ describe('DotEditContentViewAsToolbarComponent', () => {
                     lockedBy: '123'
                 }
             });
+            spyOn(dotLicenseService, 'isEnterprise').and.returnValue(of(false));
```
freddyucv: Why create a test just for this case?, I think is better add it into this one:

https://github.com/dotCMS/core-web/blob/8809f5c7bf995cd7e9810c529da7821ba372ebbc/src/app/portlets/dot-edit-page/content/components/dot-edit-content-view-as-toolbar/dot-edit-content-view-as-toolbar.component.spec.ts#L129

src/app/portlets/dot-edit-page/content/components/dot-edit-content-view-as-toolbar/dot-edit-content-view-as-toolbar.component.spec.ts
```diff
@@ -235,6 +236,7 @@ describe('DotEditContentViewAsToolbarComponent', () => {
                     JSON.parse(JSON.stringify(mockDotRenderedPage)),
                     PageMode.PREVIEW
                 );
+                spyOn(dotLicenseService, 'isEnterprise').and.returnValue(of(true));
                 fixtureHost.detectChanges();
 
                 whatsChanged = de.query(By.css('p-checkbox'));
```
freddyucv: I think we should have only one test here that check what component should be not null in case of EE license, right now we have tree test for it, also the new test should check the 'what change' checkbox

src/app/portlets/dot-edit-page/content/components/dot-edit-content-view-as-toolbar/dot-edit-content-view-as-toolbar.component.spec.ts
alfredo-dotcms: @freddyucv  this already exist in the past, I just fixed cause it was broken. I'll see if it can be done

PR #695, opened 2018-06-11
dotCMS/core#14574 Changes to content on a page not reflected on re-edit

fmontes:


src/app/portlets/dot-edit-page/content/services/dot-edit-content-html/dot-edit-content-html.service.ts
```diff
@@ -480,6 +466,10 @@ export class DotEditContentHtmlService {
         return contentletEventsMap[event];
     }
 
+    private shouldUpdateContentletInode(target: HTMLElement) {
+        return target.dataset.dotObject === 'edit-content' && target.tagName === 'BUTTON';
+    }
+
```
freddyucv: why do we need this check?

src/app/portlets/dot-edit-page/content/services/dot-edit-content-html/dot-edit-content-html.service.ts
```diff
@@ -192,17 +193,6 @@ export class DotEditContentHtmlService {
         this.currentAction = DotContentletAction.ADD;
     }
 
-    /**
-     * Set the container id where a contentlet will be update
-     *
-     * @param {string} identifier
-     * @memberof DotEditContentHtmlService
-     */
-    setContainterToEditContentlet(pageContainer: DotPageContainer): void {
```
freddyucv: COOL!!!!

src/app/portlets/dot-edit-page/content/services/dot-edit-content-html/dot-edit-content-html.service.ts
fmontes: Becuase we only need to update the contentlet inode in the DOM when we're editing from the pencil edit button at the top, not if we edit the vtl or the shortcut edit in a listing like in news.

freddyucv: but  the value of dotObject is 'contentlet' in that case, isn't it?

fmontes: `target` will be always a button to edit.

PR #726, opened 2018-07-11
Add fade-in

alfredo-dotcms:


src/app/view/components/dot-iframe-dialog/dot-iframe-dialog.component.ts
```diff
@@ -38,6 +39,7 @@ export class DotIframeDialogComponent implements OnChanges {
 
         if (changes.header) {
             this.header = changes.header.currentValue;
+            this.headerCssClass = this.header.length > 0 ? 'ui-dialog-titlebar-animate' : '';
```
freddyucv: I think we can do this directly into the template (I think is better)

src/app/view/components/dot-iframe-dialog/dot-iframe-dialog.component.scss
```diff
@@ -2,6 +2,12 @@
 @import "variables";
 
 .dialog ::ng-deep {
+    .ui-dialog.ui-dialog-titlebar-animate {
+        .ui-dialog-titlebar {
+            transition: opacity 2s ease-in;
```
fmontes: 2s is too much for this animation... the animation should be triggered when the text appear, there is a var in the scss for the animation speed.

PR #737, opened 2018-07-17
Fix bug: unable to change back to the default view

alfredo-dotcms:


alfredo-dotcms:
@freddyucv test added!

PR #745, opened 2018-07-24
issue-14868-revert-redirect-site-change

alfredo-dotcms:


src/app/view/components/dot-toolbar/dot-toolbar.component.ts
```diff
@@ -33,7 +35,9 @@ export class ToolbarComponent implements OnInit {
 
     siteChange(site: Site): void {
         this.siteService.switchSite(site);
-        this.dotNavigationService.goToFirstPortlet();
+        if (this.route.snapshot['_routerState'].url.indexOf('edit-page/') >= 0) {
+            this.dotNavigationService.goToFirstPortlet();
```
fmontes: What we talked is that it should be to `site-browser` and if anything happen, lilke the user doesn't have access to the site browser the router will redirect to first portlet automatically.

src/app/view/components/dot-toolbar/dot-toolbar.component.ts
```diff
@@ -33,7 +35,9 @@ export class ToolbarComponent implements OnInit {
 
     siteChange(site: Site): void {
```
fmontes: So if I'm in "site browser" and I'm seeing like 20 folders and I change the site, do the site browser update to reflect the folders for the new site? (the same goes with all the portlets)

alfredo-dotcms: yes it is updated

src/app/view/components/dot-toolbar/dot-toolbar.component.ts
```diff
@@ -33,7 +33,9 @@ export class ToolbarComponent implements OnInit {
...
         this.siteService.switchSite(site);
-        this.dotNavigationService.goToFirstPortlet();
+        if (this.route.snapshot['_routerState'].url.indexOf('edit-page/') >= 0) {
```
fmontes: I think we should create a method in the `dotRouterService` like `isEditPage()` to do this. @freddyucv @hmoreras?

hmoreras: sounds like it will be reusable. agree.

PR #753, opened 2018-07-31
What changed feature works only for default language

alfredo-dotcms:


src/app/portlets/dot-edit-page/content/components/dot-whats-changed/dot-whats-changed.component.spec.ts
```diff
@@ -41,6 +41,7 @@ describe('DotWhatsChangedComponent', () => {
         de = fixture.debugElement;
 
         component.pageId = '123';
+        component.languageId = '1';
```
fmontes: We need to test the ur we set here with the param because is hardcoded to `1`

PR #754, opened 2018-08-01
reloading page when a reload-edit-mode-page is tiggered

freddyucv:


src/index.html
```diff
@@ -86,10 +95,20 @@
             }
         }
     </style>
+
+    <script type="text/javascript">
+        if (window.self !== window.top) {
```
fmontes: If the page is loading in an iframe, we trigger this event.

PR #762, opened 2018-08-08
Issue 15034 bring in material design icons

alfredo-dotcms:


src/app/portlets/content-types/content-types.module.ts
```diff
@@ -103,6 +105,8 @@ import { DotBaseTypeSelectorModule } from '../../view/components/dot-base-type-s
         ConfirmDialogModule,
         ContentTypesRoutingModule,
         DialogModule,
+        DotIconModule,
```
fmontes: Do we need it here if we already have it in the `app.module`?

src/app/view/components/_common/dot-icon-button/dot-icon-button.component.ts
```diff
@@ -0,0 +1,45 @@
...
+ * The DotIconButtonComponent is a round button which
+ * reuses the dot-icon component
+ * @export
+ * @class DotIconButtonComponent
+ */
+@Component({
+    selector: 'dot-icon-button',
+    styleUrls: ['./dot-icon-button.component.scss'],
+
+    template: `<div class="dot-icon-button
```
fmontes: This seems to complex to not being in their own template file.

The click event should be in a `<button>` not a `div`, semantics matter.

src/app/view/components/_common/dot-icon-button/dot-icon-button.component.ts
```diff
@@ -0,0 +1,45 @@
...
+                {{ this.tiny ? 'tiny' : '' }}
+                {{ this.disabled ? 'disabled' : '' }}"
+                (click)='buttonOnClick($event)'>
+                   <dot-icon name="{{ this.icon }}" tiny="{{ this.tiny }}"  invertColor="{{ this.invertColor }}"></dot-icon>
+               </div>`
+})
+export class DotIconButtonComponent {
+
+    @Input() disabled?: boolean;
+    @Input() invertColor?: boolean;
```
fmontes: just inverted should be fine.

src/app/view/components/_common/dot-icon-button/dot-icon-button.component.scss
```diff
@@ -0,0 +1,30 @@
+@import 'primeng-theme/utils/theme-variables';
+
+.dot-icon-button {
```
fmontes: Remember you can style the `host`

src/app/view/components/_common/dot-icon/dot-icon.component.scss
```diff
@@ -0,0 +1,15 @@
+@import 'variables';
+
+.material-icons {
+    color: rgba(126, 122, 134, 1);
+    font-size: $font-size-xx-large;
```
fmontes: Shouldn't we create a font size variable for the icons? or just set it here directly?

src/app/view/components/_common/dot-icon/dot-icon.component.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * The DotIconComponent uses google material design icons
+ * https://material.io/tools/icons
+ * @export
+ * @class DotIconComponent
+ */
+@Component({
+    selector: 'dot-icon',
+    styleUrls: ['./dot-icon.component.scss'],
+    template: `<i class="material-icons {{ this.tiny ? 'tiny' : '' }} {{ this.invertColor ? 'invert-color' : '' }}">
```
fmontes: Same as before and https://angular.io/api/common/NgClass

src/app/portlets/content-types/content-types.module.ts
alfredo-dotcms: yes it is needed

src/app/view/components/_common/dot-icon-button/dot-icon-button.component.scss
```diff
@@ -1,30 +1,35 @@
 @import 'primeng-theme/utils/theme-variables';
+:host {
+    .dot-icon-button {
```
fmontes: You don't need a class here, even when it sound crazy, with the angular encapsulation you can do just: `button {}` because you only have one button in the component.

src/app/view/components/_common/dot-icon-button/dot-icon-button.component.scss
```diff
@@ -1,30 +1,35 @@
...
-    }
+        &:hover {
+            background-color: $gray-bg;
+        }
 
-    &.disabled:hover {
-        background-color: transparent;
-        cursor: not-allowed;
-    }
+        &.disabled:hover {
```
fmontes: Buttons have native `:disable` state.

src/app/view/components/_common/dot-icon-button/dot-icon-button.component.spec.ts
```diff
@@ -24,15 +24,15 @@ describe('DotIconButtonComponent', () => {
     it('should render component with dot-icon component and css classes', () => {
         comp.icon = 'test';
         comp.tiny = true;
-        comp.invertColor = true;
+        comp.inverted = true;
```
fmontes: I was thinking about this (because we might use it in other place) I think we don't need this as an `@Input()`

I think we can do it with css like:

```scss
:host {
    &[inverted] {
        button {
            // ...
        }

    }
}
```

I think it worth to try and if it works we should do it in any other component and we save an `Input()` that we only use for styling.

** Same goes for tiny attr **

PR #769, opened 2018-08-14
dotCMS/core#15092 Suport fa and md icons

fmontes:


src/app/view/components/_common/accordion/accordion.ts
```diff
@@ -93,6 +94,10 @@ export class AccordionGroupComponent implements AfterViewInit, OnDestroy {
         return this._isOpen;
     }
 
+    isFaIcon(icon: string): boolean {
```
alfredo-dotcms: should this be part of a util lib? where other components can use it?

fmontes: Will be deprecated.

PR #776, opened 2018-08-17
dotCMS/core#14999 Don't remove the field until the response from the server is back

fmontes:


src/app/portlets/content-types/fields/content-type-fields-row/content-type-fields-row.component.ts
```diff
@@ -89,16 +88,6 @@ export class ContentTypeFieldsRowComponent extends BaseComponent {
         }
     }
 
-    private getField(field: ContentTypeField): any {
```
alfredo-dotcms: return void

PR #780, opened 2018-08-21
We need to show the system workflow in the WF field when you are creating a new content type in community edition

alfredo-dotcms:


src/app/view/components/_common/dot-workflows-selector-field/dot-workflows-selector-field.component.scss
```diff
@@ -10,4 +10,11 @@
     color: $gray;
     font-size: $font-size-small;
     margin-left: $basic-padding;
+}
+
+:host ::ng-deep {
```
fmontes: Shouldn't this be global? like all the multiselects acts the same?

PR #781, opened 2018-08-23
dotCMS/core#15179 Bring columns back to edit content and content types

fmontes:


src/app/portlets/content-types/fields/content-type-fields-row/content-type-fields-row.component.scss
```diff
@@ -58,18 +58,22 @@
 }
 
 .row-header {
-    margin-bottom: $basic-padding * 2;
     opacity: 0.25;
     text-align: center;
     transition: opacity $basic-speed;
+    margin: 0 -#{$basic-padding * 2}
```
alfredo-dotcms: What this does? `0 -#{$basic-padding}` ?

fmontes: The box that represent a row have a `16px` padding, not instead of haveing a little icon to drag and drop a row we have the whole top section, so I add an element that goes from left to right in the top but because of the padding it was missing a little bit of space, this negative margin is to fill that gap.

PR #785, opened 2018-08-28
handle forbidden error when try to delete a content type

freddyucv:


src/app/portlets/content-types/main/content-types.component.ts
```diff
@@ -225,9 +228,13 @@ export class ContentTypesPortletComponent implements OnInit {
...
 
     private removeContentType(item): void {
-        this.crudService.delete(`v1/contenttype/id`, item.id).subscribe(() => {
-            this.listing.loadCurrentPage();
-        });
+        this.crudService.delete(`v1/contenttype/id`, item.id)
+            .pipe(
+                catchError((error) => this.httpErrorManagerService.handle(error))
+            )
+            .subscribe(() => {
```
fmontes: Does anybody knows the difference between use `catchError` or catch the error in the second param of the `subscribe`?

PR #803, opened 2018-09-20
Unable to rename category fields in the new content type portlet

alfredo-dotcms:


src/app/portlets/content-types/fields/content-type-fields-properties-form/field-properties/categories-property/categories-property.component.ts
```diff
@@ -54,6 +58,15 @@ export class CategoriesPropertyComponent extends BaseComponent implements OnInit
         this.getCategoriesList(event.filter, event.first);
     }
 
+    /**
+     * Propagate the inode value to the parent form
+     * @param {any} categoryOptionSelected
+     * @memberof CategoriesPropertyComponent
+     */
+    categoryChanged(categoryOptionSelected: any): void {
```
fmontes: Any?

src/app/portlets/content-types/fields/content-type-fields-properties-form/field-properties/categories-property/categories-property.component.ts
```diff
@@ -34,6 +36,8 @@ export class CategoriesPropertyComponent extends BaseComponent implements OnInit
             ? this.dotMessageService.get('contenttypes.field.properties.category.label')
             : this.property.value;
             this.paginationService.url = 'v1/categories';
+        this.groupLocalCopy = _.cloneDeep(this.group);
```
fmontes: Why? what's the problem here?

freddyucv: why do we need this copy?

src/app/portlets/content-types/fields/content-type-fields-properties-form/field-properties/categories-property/categories-property.component.html
```diff
@@ -1,7 +1,8 @@
-<div [formGroup]="group">
+<div [formGroup]="groupLocalCopy">
     <label class="form__label">{{ i18nMessages['categories'] }}</label>
     <dot-searchable-dropdown #searchableDropdown 
         [data]="categoriesCurrentPage"
+        (change)="categoryChanged($event)"
```
freddyucv: why the 'formControlName' doesn't work?

src/app/portlets/content-types/fields/content-type-fields-properties-form/field-properties/categories-property/categories-property.component.ts
alfredo-dotcms: at this moment `categories` comes with a categories object (categoryName, description, inode, key, keywords, sortOrder) and to save it correctly we need on that property only the `inode`.

But we also need to pass that same categories object to the `dot-searchable-dropdown`, so I made a localcopy, to pass it to the dropdown and then update the parent categories attribute with the inode, so once it gets saved the endpoint doesn't break

src/app/portlets/content-types/fields/content-type-fields-properties-form/content-type-fields-properties-form.component.ts
```diff
@@ -97,6 +97,21 @@ export class ContentTypeFieldsPropertiesFormComponent extends BaseComponent impl
         });
     }
 
+    private formatFormData(): any {
+        const formCopy = Object.assign({}, this.form.value);
```
fmontes: If this object `this.form.value` have a nested object or collection this approach will NOT clone that, the reference to the nested object doesn't change, I think this is the case here, right?

PR #813, opened 2018-10-09
issue-12667-add-fields-variables-tab-fixes

alfredo-dotcms:


src/app/portlets/content-types/fields/content-type-fields-variables/content-type-fields-variables.component.html
```diff
@@ -9,8 +9,8 @@
     <p-column field="value" [header]="messages['contenttypes.field.variables.value_header.label']" [editable]="true"></p-column>
     <p-column styleClass="content-type-fields__variables-actions" [header]="messages['contenttypes.field.variables.actions_header.label']" >
         <ng-template let-fieldVar="rowData" let-index="rowIndex" pTemplate="body">
-            <dot-icon-button icon="save" (click)="saveVariable(fieldVar, index)" [attr.disabled]="canSaveField(index) ? true : null"></dot-icon-button>
-            <dot-icon-button icon="delete" (click)="deleteVariable(index)"></dot-icon-button>
+            <dot-icon-button icon="delete_outline" (click)="deleteVariable(index)"></dot-icon-button>
+            <dot-icon-button icon="edit" (click)="saveVariable(fieldVar, index)" [attr.disabled]="canSaveField(index) ? true : null"></dot-icon-button>
```
fmontes: ```canSaveField(index)``` should't this method return `true` or `false`? also is bad to add logic to the templates, check 16: https://medium.freecodecamp.org/best-practices-for-a-clean-and-performant-angular-application-288e7b39eb6f

alfredo-dotcms: @fmontes  that method indeed return true or false. 

But this is a _special case_ `[attr.disabled]="canSaveField(index) ? true : null"` and on *false* we need to return `null` otherwise it will render disabled="false" witch will be interprete it as `true`

alfredo-dotcms: to remove it from the template, I guess this can be changed:

```canSaveField(index: number): any {
        return (this.fieldVariables[index].key === '' || this.fieldVariables[index].value === '') ? true : null;
    }```

but now it will return `any` and not a typed value

Comment (fmontes) [reply src/app/portlets/content-types/fields/content-type-fields-variables/content-type-fields-variables.component.html]: Can you just set the value to a variable and pass the variable to the attr?

Comment (alfredo-dotcms) [reply src/app/portlets/content-types/fields/content-type-fields-variables/content-type-fields-variables.component.html]: not sure how, without overcomplicating the code

Comment (alfredo-dotcms) [reply src/app/portlets/content-types/fields/content-type-fields-variables/content-type-fields-variables.component.html]: this is what I ended up with:
```
canSaveField(index: number): boolean | null {
        return this.fieldVariables[index].key === '' || this.fieldVariables[index].value === '';	        return (this.fieldVariables[index].key === '' || this.fieldVariables[index].value === '') ? true : null;
    }	    }
```

**Comment by fmontes** [comment]
I'm working on the dialogs and one problem with this dialog is that when user start adding a bunch of field variables the dialog keeps growing until is out of the screen, I'm setting a `max-height` for the dialog, but now the problem is that at some point when the user add field seems like nothing change because the fields are being added at the end of the table, we should add new field to the top of the table and push down all field variables.

File: src/app/portlets/content-types/fields/content-type-fields-variables/content-type-fields-variables.component.spec.ts [hunk 2]
```diff
@@ -77,10 +77,11 @@ describe('ContentTypeFieldsVariablesComponent', () => {
         fixture.detectChanges();
         const dataTable = de.query(By.css('p-dataTable'));
         expect(dataTable.componentInstance.value).toEqual(mockFieldVariables);
+        expect(dataTable.listeners[0].name).toBe('onEditComplete');
     });
 
-    it('should load the component and create', () => {
-        spyOn(fieldVariableService, 'save');
+    it('should load the component and create', fakeAsync(() => {
```
fmontes: Why do we need fakeAsync here?

PR #822, opened 2018-10-20
dotCMS/core#15398 Dialogs: QA feedback

fmontes:


src/app/view/components/dot-iframe-dialog/dot-iframe-dialog.component.html
```diff
@@ -1,6 +1,7 @@
 <dot-dialog
+    width="90vw"
+    height="90vh"
```
alfredo-dotcms: why?
image

fmontes: Whe I was passing the values here, the whole iframe height was: 90vh meaning that the content of the dialog it was going to be 90vh, but the dialog have header (90px) and content... that was why there was a scroll, now you can pass a height to the dialog and it will take that height (header + content)

PR #827, opened 2018-10-29
prod build happening with fake login

hmoreras:


projects/dot-rules/src/lib/app.component.ts
```diff
@@ -1,20 +1,22 @@
...
 export class AppRulesComponent implements OnInit {
     constructor(public loginService: LoginService) {}
 
     ngOnInit() {
         // TODO: need to find a better wat to login
         if (!this.loginService.auth) {
-            this.loginService.loginUser('admin@dotcms.com', 'admin', false, 'en-US').subscribe(res => {
-                console.log('fake login', res);
-            });
+            this.loginService
```
fmontes: I don't we need this login anymore.

PR #856, opened 2018-11-21
dotCMS/core#15509 Upgrade the dot-page-selector.component

fmontes:


src/app/view/components/_common/dot-page-selector/dot-page-selector.component.ts
```diff
@@ -104,4 +139,42 @@ export class DotPageSelectorComponent implements ControlValueAccessor {
...
+
+    private autoFillField($event: KeyboardEvent): void {
+        const input: HTMLInputElement = <HTMLInputElement>$event.target;
+
+        if (this.results.type === 'site') {
+            const host = <Site>this.results.data[0].payload;
+            input.value = `//${host.hostname}/`;
+        } else if (this.results.type === 'page') {
+            const page = <DotPageAsset>this.results.data[0].payload;
+            input.value = `//demo.dotcms.com${page.path}`;
```
freddyucv: this value 'demo.dotcms.com', should not come from the Backend?

src/app/view/components/_common/dot-page-selector/service/dot-page-selector.service.ts
```diff
@@ -29,59 +29,217 @@ export interface DotPageAsset {
...
 }
 
+export interface DotPageSeletorItem {
+    label: string;
+    payload: DotPageAsset | Site;
+}
+
+export interface DotPageSelectorResults {
+    data: DotPageSeletorItem[];
+    type: string;
```
freddyucv: type should be a enum, should this interface be in its own file?

src/app/view/components/_common/dot-page-selector/dot-page-selector.component.ts
```diff
@@ -23,19 +29,18 @@ import { take } from 'rxjs/operators';
...
+
+    results: DotPageSelectorResults = {
+        data: [],
+        query: '',
+        type: ''
+    };
+    val: DotPageSeletorItem;
+    message: string;
 
     constructor(private dotPageSelectorService: DotPageSelectorService) {}
```
freddyucv: maybe this service should be into this components's providers, dotPageSelectorService is a state full service so we have to be carefully with its scope

src/app/view/components/_common/dot-page-selector/service/dot-page-selector.service.ts
```diff
@@ -29,59 +29,217 @@ export interface DotPageAsset {
...
+        return [query.replace(/\//g, '\\/')];
+    }
+
+    private shouldSearchPages(param: string): boolean {
+        if (this.isReSearchingForHost(param)) {
+            this.currentHost = null;
+        }
+
+        return !!(this.currentHost || !param.startsWith('//'));
+    }
```
freddyucv: maybe this service have a lot of private util method, I don't know if a better idea create a util class and move some of the private method to there

src/app/view/components/_common/dot-page-selector/dot-page-selector.component.ts
```diff
@@ -69,11 +96,19 @@ export class DotPageSelectorComponent implements ControlValueAccessor {
...
+                .pipe(take(1))
+                .subscribe((results: DotPageSelectorResults) => {
+                    this.results = results;
+                    this.message = this.getErrorMessage();
+
+                    if (this.shouldAutoFill() && this.isOnlyFullHost(param)) {
+                        this.autoSelectUniqueResult();
+                    }
+                });
+        }
```
freddyucv: don't we need a else here?, I mean what happen if I enter something into the input and then erase it

src/app/view/components/_common/dot-page-selector/dot-page-selector.component.ts
```diff
@@ -57,9 +62,31 @@ export class DotPageSelectorComponent implements ControlValueAccessor {
...
      * @memberof DotPageSelectorComponent
      */
-    onSelect(item: DotPageAsset): void {
-        this.selected.emit(item);
-        this.propagateChange(item.identifier);
+    onSelect(item: DotPageSeletorItem): void {
+        if (this.results.type === 'site') {
+            const site: Site = <Site>item.payload;
+            this.dotPageSelectorService.setCurrentHost(site);
+            this.resetResults();
```
freddyucv: don't we need do a propagateChange here?, maybe something like: this.propagateChange(null);

I don't know if this one is the same:
https://github.com/dotCMS/core-web/pull/856/files#diff-0913f18fc1fe29a460899c0712b774b5R111

PR #914, opened 2019-01-22
dotCMS/core#15339 Add tab field feature to new content type 

hmoreras:


src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
```diff
@@ -24,12 +24,48 @@ export class DotMaxlengthDirective {
...
-    eventHandler() {
-        setTimeout(() => {
-            this.el.nativeElement.textContent = this.el.nativeElement.textContent.slice(
-                0,
-                this._maxLength
-            );
-        }, 0);
+    eventHandler(event: Event) {
+        if (
+            this.el.nativeElement.textContent.length < this._maxLength ||
```
fmontes: This smells, first you do:

`this.el.nativeElement.textContent.length < this._maxLength`

and then inside yo do the opposite: `this.el.nativeElement.textContent.length > this._maxLength`

are you sure about this?

fmontes: This smells, first you do:

this.el.nativeElement.textContent.length < this._maxLength

and then inside yo do the opposite: this.el.nativeElement.textContent.length > this._maxLength

are you sure about this?

hmoreras: When the event fires, the textContent have the initial value, with the timeout the value is updated. example, on a key press event of "B" the before the timeout value will be "A", after will be "AB".

src/app/view/directives/dot-maxlength/dot-maxlength.directive.spec.ts
```diff
@@ -20,28 +19,42 @@ describe('DotMaxlengthDirective', () => {
...
-        element.triggerEventHandler('keypress', 'Longer text example');
-        expect(directiveInstance.eventHandler).toHaveBeenCalled();
-    });
+    it(
+        'should allow new characters when max length is not reached',
+        fakeAsync(() => {
+            const event = new Event('keypress');
+            fixture.detectChanges();
+            element.nativeElement.dispatchEvent(event);
+            element.nativeElement.textContent = element.nativeElement.textContent + 'o';
```
fmontes: Here looks like you're doing what the directive should do it it self... are you sure about this test?

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
```diff
@@ -7,29 +9,78 @@ import { Directive, ElementRef, HostListener, Input } from '@angular/core';
...
-        }, 0);
+    isAllowedKeyCode(event) {
+        return (
+            event.keyCode === 8 ||
+            event.keyCode === 38 ||
+            event.keyCode === 39 ||
+            event.keyCode === 37 ||
+            event.keyCode === 40 ||
+            event.ctrlKey ||
+            event.metaKey
```
freddyucv: maybe we should create a enum here and do someyhing like:

          event.keyCode === KEY.BACKSPACE ||
           event.keyCode === KEY.LEFT_KEY ||
           event.keyCode === KEY.UP_KEY ||
           ...

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
```diff
@@ -7,29 +9,78 @@ import { Directive, ElementRef, HostListener, Input } from '@angular/core';
...
+export class DotMaxlengthDirective implements OnInit, OnDestroy {
     private _maxLength: number;
+    private events = ['paste', 'keypress'];
+    private destroy$: Subject<boolean> = new Subject<boolean>();
 
     constructor(private el: ElementRef) {}
 
+    ngOnInit() {
+        const eventStreams = this.events.map(ev => fromEvent(this.el.nativeElement, ev));
+        const allEvents$ = merge(...eventStreams);
```
fmontes: Why the `...` here?

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
```diff
@@ -7,29 +9,78 @@ import { Directive, ElementRef, HostListener, Input } from '@angular/core';
...
+            .pipe(
+                takeUntil(this.destroy$),
+                tap((keyboardEvent: Event) => {
+                    if (!this.isValidAction(keyboardEvent)) {
+                        keyboardEvent.preventDefault();
+                    }
+                }),
+                delay(1),
+                filter(() => this.el.nativeElement.textContent.length > this._maxLength),
+                tap(() => this.reduceText())
```
fmontes: Why here and not in the subscribe?

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
```diff
@@ -7,29 +9,78 @@ import { Directive, ElementRef, HostListener, Input } from '@angular/core';
...
-    @HostListener('paste', ['$event'])
-    @HostListener('keypress', ['$event'])
-    eventHandler() {
-        setTimeout(() => {
-            this.el.nativeElement.textContent = this.el.nativeElement.textContent.slice(
-                0,
-                this._maxLength
-            );
-        }, 0);
+    isAllowedKeyCode(event) {
```
fmontes: Type?

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
```diff
@@ -7,29 +9,78 @@ import { Directive, ElementRef, HostListener, Input } from '@angular/core';
...
+        return (
+            event.keyCode === 8 ||
```
fmontes: There is a property of the Keyboard event that is the string of the so you don't have to use a number and specify what it is in the doc.

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
```diff
@@ -7,29 +9,78 @@ import { Directive, ElementRef, HostListener, Input } from '@angular/core';
...
+        this.destroy$.next(true);
+        this.destroy$.complete();
+    }
+
     @Input()
     set dotMaxlength(maxLength: number) {
         this._maxLength = maxLength || 255;
     }
 
+    private isValidAction(keyboardEvent: Event): boolean {
```
fmontes: Are you sure this is an `Event` and not a `KeyboardEvent`? and if it it both (because of the paste) you can do `keyboardEvent: Event | KeyboardEvent`.

src/app/view/directives/dot-maxlength/dot-maxlength.directive.spec.ts
```diff
@@ -20,28 +19,42 @@ describe('DotMaxlengthDirective', () => {
...
 
     it(
-        'should remove extra characters',
+        'should remove extra characters when length is more than max length',
         fakeAsync(() => {
+            const event = new Event('paste');
             fixture.detectChanges();
-            directiveInstance.eventHandler();
+            element.nativeElement.dispatchEvent(event);
+            element.nativeElement.textContent = '12345678901';
```
fmontes: And the same here... shouldn't the `textContent` be updated with the event dispatched?

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
```diff
@@ -7,29 +9,78 @@ import { Directive, ElementRef, HostListener, Input } from '@angular/core';
...
+        return (
+            this.el.nativeElement.textContent.length < this._maxLength ||
+            this.isAllowedKeyCode(keyboardEvent) ||
```
freddyucv: why are we checking this?

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
hmoreras: a way to do it, `export function merge<T>(...observables: (ObservableInput<T> | SchedulerLike | number)[]): Observable<T>;`

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
hmoreras: can be either, initially based on your recommendation.

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
hmoreras: because in FF, when you do any of these will fire an event, and since we are preventing default if the string is with full length, those event get blocked and we don't want that.

src/app/view/directives/dot-maxlength/dot-maxlength.directive.ts
hmoreras: changing to key string

PR #949, opened 2019-03-06
 dotCMS/core#15848 We need to show visual scroller on field properties

fmontes:


src/app/view/components/dot-dialog/dot-dialog.component.scss
```diff
@@ -35,8 +35,19 @@ $dialog-padding: $basic-padding * 4;
 .dialog__content {
     flex-basis: 0%;
     flex-grow: 1;
+    overflow-x: hidden;
```
fmontes: It was showing in Safari when the Y scroll takes space.

PR #963, opened 2019-03-28
dotCMS/core#16289 Create DotApiContentType class for handle Content Types API

alfredo-dotcms:


projects/dotcms/src/lib/models/DotCMSContentTypeField.model.ts
```diff
@@ -0,0 +1,17 @@
+export interface DotCMSContentTypeField {
```
fmontes: Please move this one inside `DotCMSContentType.model.ts` what I'm trying is to have **one model** file per API, I think it would be cleaner that way.

PR #966, opened 2019-03-29
dotCMS/core#16167 Implement redesign of login page

hmoreras:


src/app/view/components/login/main/dot-login-page.component.scss
```diff
@@ -5,6 +5,15 @@
     width: 100%;
     display: flex;
     justify-content: center;
+
+    ::ng-deep {
+        .error-message {
+            min-height: $line-height;
```
fmontes: Why here and no directly in the component?

hmoreras: don't repeat the rule 3 times.. in each component.

PR #971, opened 2019-04-02
dotCMS/core#16290 Create DotApiForm class to get and render a DotCMS form

alfredo-dotcms:


projects/dotcms/src/lib/api/DotApiForm.ts
```diff
@@ -0,0 +1,80 @@
...
+            fieldTags += config.fields.includes(field.variable) ? this.createField(field) : '';
+        });
+
+        this.formScript += `<form>${fieldTags}</form>`;
+    }
+
+    // tslint:disable-next-line:cyclomatic-complexity
+    private createField(field: DotCMSContentTypeField): string {
+        const fieldClazz = field.clazz.split('.');
+        const fieldTag = FieldElementsTags[fieldClazz[fieldClazz.length - 1]];
```
fmontes: Screen Shot 2019-04-02 at 10 09 40 AM

fmontes: Also, we can create a `getFieldTag()` method to handle all this. Is cleaner a method should do one thing.

projects/dotcms/src/lib/api/DotApiForm.ts
```diff
@@ -0,0 +1,80 @@
...
+        return fieldTag
+            ? `
+            <${fieldTag}
```
fmontes: This solution is not very scalable, what happen when the fields have different properties? Or if doesn't have one of this properties?

We need to iterate over the properties of the object and base on that create the elements attributes. We might have to create a map for like `name > label`, `defaultValue > value` and stuff like that. We might even need to create a map of this kind base on each dotcms field type, not sure.

projects/dotcms/src/lib/utils/fieldsElementsTags.ts
```diff
@@ -0,0 +1,5 @@
+export enum FieldElementsTags {
```
fmontes: I don't think we're going to reuse this anywhere else but inside the class, so we can just move it there and not export it.

projects/dotcms-field-elements/src/utils/utils.ts
```diff
@@ -6,8 +6,7 @@ export function generateId(): number {
 
 export function getItemsFromString(rawString: string): DotOption[] {
```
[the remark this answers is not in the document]
alfredo-dotcms: no

projects/dotcms/src/lib/api/tests/DotApiForm.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    beforeEach(() => {
+        dotApiContentType = new DotApiContentTypeMock();
+    });
+
+    it('should render a Form', async () => {
+        const config = { identifier: '321', fields: ['field1', 'field2'] };
+        const container = document.createElement('div');
+        dotApiForm = new DotApiForm(dotApiContentType, config);
+        dotApiForm.render(container).then(() => {
+            expect(container.innerHTML.includes('<script type="module">')).toEqual(true);
```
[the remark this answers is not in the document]
fmontes: Also... we should not only validate the inner HTML we should validate that it used `append` to add the content to the container because is someone remove the append and does `innerHTML = ` it breaks the functionality but the tests will keep passing.

projects/dotcms/src/public_api.ts
```diff
@@ -39,11 +41,17 @@ export const initDotCMS = (config: DotCMSConfigurationParams): DotCMSApp => {
         contentType: new DotApiContentType(httpClient),
         esSearch: new DotApiElasticSearch(httpClient),
         event: new DotApiEvent(),
+        form: {
+            get: (formConfig: DotCMSFormConfig) => {
+                const dotApiContentType = new DotApiContentType(httpClient);
+                return new DotApiForm(dotApiContentType, formConfig);
```
fmontes: Yup... this makes so muuuuuch more sense!

projects/dotcms/src/lib/api/DotApiForm.ts
```diff
@@ -0,0 +1,91 @@
...
+     * @memberof DotApiForm
+     */
+    async render(container: HTMLElement) {
+        this.fields = this.fields
+            ? this.fields
+            : await this.dotApiContentType.getFields(this.formConfig.identifier);
+
+        const formScript = this.createForm(this.fields);
+        const importScript = document.createElement('script');
+        const formTag = document.createElement('div');
```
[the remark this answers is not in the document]
alfredo-dotcms: right now the `div` it's a wrapper, but probably that will be changed on the next issue/task

projects/dotcms/src/lib/api/tests/DotApiForm.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { DotApiForm } from '../DotApiForm';
+import { DotCMSContentTypeField } from 'dotcms/lib/models';
+
+function validateForm(formScript: string): void {
```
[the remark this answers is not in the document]
alfredo-dotcms: the whole string contains a lot of break lines and spaces that makes it hard to compare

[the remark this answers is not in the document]
fmontes: This `it` have 8 expect, is too much we had issues for track down broken test because of this in the pass.

In this case here is you can do:

`console.log(container.innerHTML)`

Copy what is there and then you can do:

```
expect(container.innerHTML).toBe(`
<h1>    Hello</h1>
    <label>stuff</label>
<input type="   check">
`)
```

And that way any string no matter spaces or break likes will be expect correctly in one expect... I think 😅

projects/dotcms/src/lib/api/DotApiForm.ts
freddyucv: we should have a util class for this with a map like:

```
map = {
    name: `label="${field.name}"`,
    defaultValue: `value="${field.defaultValue}"`
    ...
}

...

getProperties(field: DotCMSContentTypeField) {
    result: string = '';
    for (property in field) {
        result += property;
    }
}
```

projects/dotcms/src/lib/api/tests/DotApiForm.spec.ts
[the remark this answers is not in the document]
freddyucv: you could remove them before compare

projects/dotcms/src/lib/models/DotCMSContentType.model.ts
```diff
@@ -13,6 +15,7 @@ export interface DotCMSContentTypeField {
     dataType?: string;
     hint?: string;
     fieldTypeLabel?: string;
+    values?: string;
     variable?: string;
 }
```
freddyucv: can we use here the interface define in angular?

projects/dotcms/src/lib/models/DotCMSForm.model.ts
```diff
@@ -0,0 +1,8 @@
+export interface DotCMSFormConfig {
```
freddyucv: doc

projects/dotcms/src/lib/models/DotCMSContentType.model.ts
alfredo-dotcms: @fmontes what do you think?

fmontes: Even tho I think we should try to unify types somehow, in this case this is not a good approach because what we have in angular and this lib are completely different projects.

PR #972, opened 2019-04-02
dotCMS/core#16314 Align dot-field-helper and fix the overflow of the field text

hmoreras:


src/app/view/components/_common/dot-page-selector/dot-page-selector.component.scss
```diff
@@ -9,13 +10,14 @@
         .ui-autocomplete-list-item {
             white-space: nowrap;
         }
+
+        .autocomplete-input {
+            padding-right: $basic-padding * 4;
```
fmontes: Why don't we create a global class that we can put in any input to do this? so in case we have yo use the helper in another input, we just use that clase.

src/styles/_forms.scss
```diff
@@ -43,6 +44,16 @@
     margin: 6px 0 0;
 }
 
+.form__group--helper {
```
fmontes: Shouldn't we use this class in the component instead of adding specific css properties there?

hmoreras: this is for inputs in specific, and also needed to set the `dot-field-helper` position. better here to keep styles related to  helper in a single place ?

src/app/view/components/_common/dot-page-selector/dot-page-selector.component.scss
```diff
@@ -9,13 +10,14 @@
         .ui-autocomplete-list-item {
             white-space: nowrap;
         }
+
+        .ui-autocomplete-input {
+            padding-right: $basic-padding * 4 !important;
```
hmoreras: needed because the `p-autocomplete` component can add inline styles.

PR #994, opened 2019-04-22
Custom Elements DRY

fmontes:


projects/dotcms-field-elements/src/components/dot-textfield-1/dot-textfield-1.tsx
```diff
@@ -0,0 +1,187 @@
...
+};
+
+const getTagLabel = (name: string, label: string): string => {
+    return <label htmlFor={name}>{label}</label>;
+};
+
+const getErrorClass = (valid: boolean): string => {
+    return valid ? '' : 'dot-field__input--error';
+};
+
```
alfredo-dotcms: AFAIK, from line 1 to here, all this will go to the `Utils` class?

if so, prob we can remove `@Event() valueChange: EventEmitter<DotFieldValueEvent>;` and `@Event() statusChange: EventEmitter<DotFieldStatusEvent>;` from each field and also the `emits`, so in the end the `util` class will have a method to emit from there

fmontes: But how do we pass the `this.statusChange` and `this.valueChange` to that function?

alfredo-dotcms: I guess you can pass the `JSON Object` (based on type), and then the util function will only do the emit of the JS object

PR #1052, opened 2019-06-05
Dot date test

hmoreras:


projects/dotcms-field-elements/src/components/dot-date/dot-date.e2e.ts
```diff
@@ -1,131 +1,385 @@
...
 
-    it('should be valid, touched & dirty on value change', async () => {
-        await input.press('2');
-        await page.waitForChanges();
-        expect(element.classList.contains('dot-valid')).toBe(true);
-        expect(element.classList.contains('dot-dirty')).toBe(true);
-        expect(element.classList.contains('dot-touched')).toBe(true);
-    });
+        it('should be valid, touched & dirty when filled', async () => {
+            inputCalendar.triggerEvent('_statusChange', {
```
fmontes: Why don't you create a function where you just pass something like `triggerStatusChange(false, true, true)` because there is a lot of repeating code here.

projects/dotcms-field-elements/src/components/dot-date/dot-date.e2e.ts
```diff
@@ -1,131 +1,385 @@
...
+
+            it('should render required attribute for the dot-label', async () => {
+                element.setProperty('required', 'true');
+                await page.waitForChanges();
+                const label = await dotTestUtil.getDotLabel(page);
+                expect(label.getAttribute('required')).toBeDefined();
+            });
+        });
+
+        describe('requiredMessage', () => {
```
fmontes: Here you're not testing that the error show, I can go to the component and remove [this](https://github.com/dotCMS/core-web/blob/f762318d7be8e8378cd8cf5416fa6d21c32521f7/projects/dotcms-field-elements/src/components/dot-date/dot-date.tsx#L144) and the test will keep passing.

In fact, passing the messages as atrr to a child component to then get them back in an event is not a good approach because we're passing down a message that will be shown where is getting passed from, so it doesn't make sense, this was something we all miss in the code review.

The error message can be calculated to show or hide using the `require` prop and the `status` in this case the `dot-input-calendar` don't have to do it itself this is a job of the parent to do that.

projects/dotcms-field-elements/src/components/dot-date/dot-date.e2e.ts
```diff
@@ -1,131 +1,385 @@
...
+            });
+
+            it('should pass correctly to dot-input-calendar', async () => {
+                element.setProperty('requiredMessage', 'test');
+                await page.waitForChanges();
+                expect(inputCalendar.getAttribute('required-message')).toBe('test');
+            });
+        });
+
+        describe('validationMessage', () => {
```
fmontes: Same as `requiredMessage` comment.

projects/dotcms-field-elements/src/components/dot-date/dot-date.e2e.ts
```diff
@@ -1,131 +1,385 @@
...
+                        dotPristine: false,
+                        dotTouched: true,
+                        dotValid: true
+                    }
+                });
+                expect(spyValueChangeEvent).toHaveReceivedEventDetail({
+                    name: '',
+                    value: '2019-01-01'
+                });
+                expect(evt_statusChange.events).toEqual([]);
```
fmontes: What are you testing here?

hmoreras: that the `stopImmediatePropagation` for `evt_statusChange` happened.

projects/dotcms-field-elements/src/components/dot-date/dot-date.e2e.ts
hmoreras: fixed with the interface, `DotInputCalendarStatusEvent `

projects/dotcms-field-elements/src/components/dot-date/dot-date.tsx
```diff
@@ -10,7 +10,11 @@ import {
     Watch
 } from '@stencil/core';
 import Fragment from 'stencil-fragment';
-import { DotFieldStatusClasses, DotFieldStatusEvent, DotFieldValueEvent } from '../../models';
+import {
+    DotFieldStatusClasses,
+    DotFieldValueEvent,
+    DotInputCalendarStatusEvent
```
fmontes: Shouldn't we move this to a `/models` folder inside `dot-input-calendar`?

projects/dotcms-field-elements/src/components/dot-date/dot-date.tsx
```diff
@@ -88,27 +100,24 @@ export class DotDateComponent {
...
+        const valueEvent: DotFieldValueEvent = event.detail;
+        this.value = valueEvent.value;
+        this.valueChange.emit(valueEvent);
     }
 
     @Listen('_statusChange')
     emitStatusChange(event: CustomEvent) {
         event.stopImmediatePropagation();
-        const statusEvent: DotFieldStatusEvent = event.detail;
+        const statusEvent: DotInputCalendarStatusEvent = event.detail;
```
fmontes: `statusEvent` could get confused with the event itself this is only caching the value of the event.

projects/dotcms-field-elements/src/components/dot-input-calendar/dot-input-calendar.tsx
```diff
@@ -1,6 +1,6 @@
 import { Component, Element, Event, EventEmitter, Method, Prop, State } from '@stencil/core';
```
fmontes: Don't we need a type default value?

projects/dotcms-field-elements/src/components/dot-date/dot-date.e2e.ts
```diff
@@ -1,131 +1,327 @@
...
+    describe('@Props', () => {
+        describe('value', () => {
+            it('should render default value', () => {
+                expect(inputCalendar.getAttribute('value')).toBe('');
+            });
+
+            it('should pass correctly to dot-input-calendar', async () => {
+                element.setProperty('value', '2019-01-01');
+                await page.waitForChanges();
+                expect(await inputCalendar.getProperty('value')).toBe('2019-01-01');
```
fmontes: getAttribute?

projects/dotcms-field-elements/src/components/dot-date/dot-date.tsx
```diff
@@ -21,46 +25,54 @@ export class DotDateComponent {
...
-    @Prop() step = '1';
+    @Prop({ reflectToAttr: true })
+    step = '1';
 
     @State() classNames: DotFieldStatusClasses;
     @State() errorMessageElement: JSX.Element;
 
     @Event() valueChange: EventEmitter<DotFieldValueEvent>;
-    @Event() statusChange: EventEmitter<DotFieldStatusEvent>;
+    @Event() statusChange: EventEmitter<DotInputCalendarStatusEvent>;
```
fmontes: Should we emit the whole thing? or should we keep emiting the same as all other field? I mean, clean turn the `DotInputCalendarStatusEvent` into `DotFieldStatusEvent` and then emit it because is what all the other fields emit.

projects/dotcms-field-elements/src/models/dot-field-event.model.ts
```diff
@@ -8,6 +8,10 @@ export interface DotFieldStatusEvent extends DotFieldEvent {
     status: DotFieldStatus;
 }
 
+export interface DotInputCalendarStatusEvent extends DotFieldStatusEvent {
+    inValidRange: boolean;
```
fmontes:
```suggestion
    isValidRange: boolean;
```

projects/dotcms-field-elements/src/utils/test/utils.tsx
```diff
@@ -12,5 +12,24 @@ export const dotTestUtil = {
         filledRequired: ['dot-required', 'dot-valid', 'dot-touched', 'dot-dirty'],
         filledRequiredPristine: ['dot-required', 'dot-valid', 'dot-untouched', 'dot-pristine'],
         touchedPristine: ['dot-valid', 'dot-pristine', 'dot-touched']
+    },
+    triggerStatusChange: (
+        pristine: boolean,
```
fmontes: too many params? maybe object?

projects/dotcms-field-elements/src/components/dot-date/dot-date.e2e.ts
hmoreras: actually I tried that, but can't explain why `getAttribute` do not work in this scenario.

projects/dotcms-field-elements/src/components/dot-date/dot-date.tsx
hmoreras: right, mi bad.

projects/dotcms-field-elements/src/components/dot-input-calendar/dot-input-calendar.tsx
hmoreras: only used internally.

projects/dotcms-field-elements/src/models/dot-field-event.model.ts
hmoreras: done

projects/dotcms-field-elements/src/components/dot-date/dot-date.tsx
hmoreras: it will be used in 4 custom fields in total.

projects/dotcms-field-elements/src/utils/test/utils.tsx
hmoreras: agree, but create an object will add more "noise" to the calls in the test.

PR #1055, opened 2019-06-06
Fix issue-fix-dot-date-range-test

alfredo-dotcms:


projects/dotcms-field-elements/src/components/dot-date-range/dot-date-range.tsx
```diff
@@ -142,8 +142,8 @@ export class DotDateRangeComponent {
             mode: 'range',
             altFormat: this.displayFormat,
             altInput: true,
-            maxDate: this.parseDate(this.max),
-            minDate: this.parseDate(this.min),
+            maxDate: this.max ? this.parseDate(this.max) : null,
+            minDate: this.min ? this.parseDate(this.min) : null,
```
freddyucv: test for this change?

projects/dotcms-field-elements/src/utils/props/validators/props.ts
```diff
@@ -28,6 +28,16 @@ export function dateValidator<T>(propInfo: PropValidationInfo<T>): void {
     }
 }
 
+export function dateRangeValidator<T>(propInfo: PropValidationInfo<T>): void {
```
freddyucv: doc?

Really this function not validate a Range date, I mean when I read 'dateRangeValidator' I think that tis one validate a Max and Min date but really this validate date format for two date, I don't now if the name is the right one here

projects/dotcms-field-elements/src/components/dot-date-range/dot-date-range.tsx
freddyucv: test?

alfredo-dotcms: no idea how to test this, I mean this is added to remove the noise that's printed on the terminal when running unit test.

With or without this, there's not a difference in the DOM, so I have no idea how I could so any assert on this

fmontes: Can you add a little range min and max, open the table, get all the <td> and check which one have a specific class?

projects/dotcms-field-elements/src/components/dot-date-range/dot-date-range.tsx
```diff
@@ -119,8 +119,8 @@ export class DotDateRangeComponent {
 
     @Watch('value')
```
fmontes: Can we add to the description an example, like:

```Value formatted with start and end date splitted with a comma, ex: 02-20-2012,01-10-2012```

The same should go for any other prop that have like a very specific format of date and stuff, cc: @hmoreras

PR #1062, opened 2019-06-13
dotCMS/core#15773 Enter key should save the bundle on the rules potlet

alfredo-dotcms:


projects/dot-rules/src/lib/push-publish/add-to-bundle-dialog-component.ts
```diff
@@ -21,8 +21,7 @@ import { MenuItem } from 'primeng/api';
         [value]="bundleStores ? bundleStores[0]?.id : null"
         allowAdditions="true"
         (onDropDownChange)="setSelectedBundle($event)"
-        (enter)="addToBundle.emit(selectedBundle)">
-
+        (keyup.enter)="addToBundle.emit(selectedBundle)">
```
fmontes: What happen if I hit enter with an empty value?

alfredo-dotcms: nothing it does not allow to add it, unless you have a value

PR #1065, opened 2019-06-18
dotCMS/core#16737 Code review feedback

fmontes:
1. Page id was not getting passed correctly for all cases
2. Make rules button disable only when no license

### DIFF
--- a/projects/dot-rules/src/lib/services/Rule.ts
+++ b/projects/dot-rules/src/lib/services/Rule.ts
@@ -240,11 +240,17 @@ export const DEFAULT_RULE: IRule = {
 // @dynamic
 @Injectable()
 export class RuleService {
+
+  get rules(): RuleModel[] {
+    return this._rules;
+  }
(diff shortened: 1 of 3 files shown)


projects/dot-rules/src/lib/services/Rule.ts
```diff
@@ -551,25 +551,28 @@ export class RuleService {
...
-      if (document.location.hash.indexOf('?') >= 0) {
-        query = document.location.hash.substr(document.location.hash.indexOf('?') + 1);
-      }
-    }
-
-    /**
-     * Search if the realId parameter is set
-     */
-    siteId = ApiRoot.parseQueryParam(query, 'realmId');
+    let siteId = this.getPageIdFromUrl();
```
fmontes: The change was this. All other stuff was just code auto aligning.

projects/dot-rules/src/lib/services/Rule.ts
```diff
@@ -551,25 +551,28 @@ export class RuleService {
     return obs;
   }
 
+  private getPageIdFromUrl(): string {
```
fmontes: new method.

PR #1072, opened 2019-06-25
dotCMS/core#16774 Update custom fields css

fmontes:


projects/dotcms-field-elements/src/components/dot-label/dot-label.scss
```diff
@@ -0,0 +1,33 @@
+.dot-field__error-message,
```
fmontes: All this css for hint and message will go away, I'm planning for future release create the hint and error as their own components.

PR #1073, opened 2019-06-25
dotCMS/core#16461 Hook dot-tag with dotcms tag api

fmontes:


projects/dotcms-field-elements/src/components/dot-form/utils/fields.tsx
```diff
@@ -120,6 +120,16 @@ export const DotFormFields = {
 
     Tag: (field: DotCMSContentTypeField) => (
         <dot-tags
+            data={async (): Promise<string[]> => {
+                return fetch('//localhost:8080/api/v1/tags', {
```
fmontes: For some reason inside the `dot-form` this gets trigger everytime we add a tag.

freddyucv: this 'localhost:8080' should not be hard code

projects/dotcms-field-elements/src/components/dot-tags/dot-tags.tsx
```diff
@@ -29,6 +23,9 @@ export class DotTagsComponent {
     /** Value formatted splitted with a comma, for example: tag-1,tag-2 */
     @Prop({ mutable: true, reflectToAttr: true }) value = '';
 
+    /** Name that will be used as ID */
+    @Prop() data: () => Promise<string[]>;
```
freddyucv: doc is wrong

projects/dotcms-field-elements/src/components/dot-tags/components/dot-autocomplete/dot-autocomplete.tsx
```diff
@@ -1,6 +1,20 @@
 import { Component, Prop, Event, EventEmitter, Element, Watch } from '@stencil/core';
 import autoComplete from '@tarekraafat/autocomplete.js/dist/js/autoComplete';
 
+interface SelectionItem {
+    index: number;
+    value: string;
+    match: string;
+}
+
+interface SelectionFeedback {
```
freddyucv: these interfaces should not come from '@tarekraafat/autocomplete.js/dist/js/autoComplete'?

projects/dotcms-field-elements/src/components/dot-tags/dot-tags.tsx
freddyucv: really, if the user need to use this component out the dot-form need to pass this functions always, I think it should have a default value maybe

projects/dotcms-field-elements/src/components/dot-tags/components/dot-autocomplete/dot-autocomplete.tsx
fmontes: No because that's not a ts package, and it doesn't have a types package.

projects/dotcms-field-elements/src/components/dot-tags/dot-tags.tsx
fmontes: No, because what we need to publish in the library is a tag field that can support autocomplete form X source, that X will be given for who is using it.

PR #1081, opened 2019-06-27
dotCMS/core#16772 When you DnD a field (reorder) in the same column it shows up the edit field dialog

alfredo-dotcms:


src/app/portlets/content-types/fields/content-type-fields-drop-zone/content-type-fields-drop-zone.component.ts
```diff
@@ -210,10 +213,12 @@ export class ContentTypeFieldsDropZoneComponent implements OnInit, OnChanges, On
      * @memberof ContentTypeFieldsDropZoneComponent
      */
     editFieldHandler(fieldToEdit: DotContentTypeField): void {
-        const fields = FieldUtil.getFieldsWithoutLayout(this.fieldRows);
-        this.currentField = fields.find((field: DotContentTypeField) => fieldToEdit.id === field.id);
-        this.currentFieldType = this.fieldPropertyService.getFieldType(this.currentField.clazz);
-        this.toggleDialog();
+        if (!this.draggedEvent) {
```
freddyucv: I think is better handle this boolean into the fieldDragDropService, and in this line you can do something like:

```suggestion
        if (!this.fieldDragDropService.draggedEventStarted()) {
```

PR #1085, opened 2019-07-11
dotCMS/core#16772 When you DnD a field (reorder) in the same column it shows up the edit field dialog v2

alfredo-dotcms:


src/app/portlets/content-types/fields/service/field-drag-drop.service.ts
```diff
@@ -72,6 +72,9 @@ export class FieldDragDropService {
         this._fieldDropFromTarget = dragulaService.dropModel().pipe(
             tap(() => {
                 this.draggedEvent = true;
+                setTimeout(() => {
+                    this.draggedEvent = false;
```
fmontes: Because this is mock in the tests, if I came here and remove this line, will the test break?

alfredo-dotcms: yes

PR #1089, opened 2019-07-15
dotCMS/core#16792 Add friendly message when there are no rules

hmoreras:


projects/dot-rules/src/lib/rule-engine.ts
```diff
@@ -46,7 +46,8 @@ const I8N_BASE = 'api.sites.ruleengine';
     </div>
   </div>
   <div class="cw-rule-engine__empty" *ngIf="!rules.length">
-      <span>{{rsrc('inputs.want.to.add.rules') | async}}</span>
+      <i class="material-icons">tune</i>
```
fmontes: We have a dot-icon component: https://github.com/dotCMS/core-web/blob/master/src/app/view/components/_common/dot-icon/dot-icon.component.html#L1

projects/dot-rules/src/lib/styles/rule-engine.scss
```diff
@@ -505,9 +507,23 @@ $finder-s-input: '.ui.input input';
...
+        i {
+            font-size: 120px;
+            color: $gray-light;
+            border: 7px solid;
+            border-radius: 10px;
+            margin-bottom: $basic-padding * 3;
+        }
+
+        h2 {
+            font-size: 22px;
```
fmontes: We have variables for this.

projects/dot-rules/src/lib/styles/rule-engine.scss
```diff
@@ -14,6 +14,8 @@ $foreground-active: #848484;
 $text-emphasis: #fc7e1c;
 $warn: #cb3927;
 $brand-primary-one: #209021;
+$gray: #7e7a86;
```
fmontes: Are this the same grays we have in the app variables?

hmoreras: yes

PR #1125, opened 2019-08-20
dotCMS/core#16840 Show Country next to language v2

alfredo-dotcms:


src/app/view/components/dot-language-selector/dot-language-selector.component.ts
```diff
@@ -40,7 +40,7 @@ export class DotLanguageSelectorComponent implements OnInit {
     private decorateLabels(languages: DotLanguage[]): any {
```
fmontes: `any`? Why if we only changing the string of the language label you need to create a new prop in the entry instead of just modifying the string of the name?

PR #1130, opened 2019-08-22
dotCMS/core#17003 Add Rule: It is not possible to add a rule with the options loaded by default

alfredo-dotcms:


projects/dot-rules/src/lib/rule-condition-component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, EventEmitter, Input, Output, OnInit } from '@angular/core';
+import { Component, EventEmitter, Input, Output, AfterViewInit, OnInit } from '@angular/core';
```
hmoreras: remove `AfterViewInit `

alfredo-dotcms: done!

PR #1154, opened 2019-09-04
dotCMS/core#17097 Modularize dot-content-types-edit component

alfredo-dotcms:


src/app/portlets/content-types/content-types.module.ts
```diff
@@ -1,167 +1,11 @@
...
-        ContentTypeEditResolver,
-        DotContentTypesInfoService,
-        DotWorkflowService,
-        DragulaService,
-        FieldDragDropService,
-        FieldPropertyService,
-        FieldService,
-        FormatDateService
-    ]
+    imports: [CommonModule, ContentTypesRoutingModule, DotListingDataTableModule],
```
fmontes: Do we need `DotListingDataTableModule` here? since we have that in the listing module?

src/app/portlets/content-types/content-types-routing.module.ts
```diff
@@ -1,7 +1,6 @@
-import { ContentTypesEditComponent } from './edit';
```
fmontes: Screen Shot 2019-09-04 at 10 38 01 AM

All this should be in shared.

PR #1167, opened 2019-09-11
dotCMS/core#17177 Make dot-binary support "maxFileLength" attribute

alfredo-dotcms:


projects/dotcms-field-elements/src/components/dot-binary-file/dot-binary-file.tsx
```diff
@@ -251,6 +259,7 @@ export class DotBinaryFileComponent {
                 </dot-label>
                 {getTagHint(this.hint)}
                 {getTagError(this.shouldShowErrorMessage(), this.getErrorMessage())}
+                {this.errorMessage ? <div class="dot-binary__error-message">{this.errorMessage}</div> : ''}
```
fmontes: Why don't we use here: `dot-form-error-message` component so we don't have to add any extra styling

PR #1191, opened 2019-09-24
Add indexPolicy to create persona form

fmontes:


src/app/view/components/dot-add-persona-dialog/dot-add-persona-dialog.component.ts
```diff
@@ -93,6 +66,29 @@ export class DotAddPersonaDialogComponent implements OnInit {
         this.dialogActions.accept.disabled = true;
     }
 
+    private savePersona(): void {
+        if (this.personaForm.form.valid) {
+            this.dotWorkflowActionsFireService
+                .publishContentlet<DotPersona>(PERSONA_CONTENT_TYPE, {
+                    ...this.personaForm.form.getRawValue(),
+                    ...{ indexPolicy: 'WAIT_FOR' }
```
freddyucv: indexPolicy always should be WAIT_FOR?, if it can has another values then it should be a enum, if always should be 'WAIT_FOR' then we should add it into 'publishContentlet' method

PR #1193, opened 2019-09-25
dotCMS/core#17178  Link content types entries to contentlets results

hmoreras:


src/app/view/components/dot-listing-data-table/dot-listing-data-table.component.html
```diff
@@ -47,7 +47,7 @@
                 <a *ngIf="col.fieldName === 'nEntries'"
                    [routerLink]="rowData.variable === 'Host' ? '/c/sites' : '/c/content'"
                    [queryParams]="rowData.variable === 'Host' ? {} : {filter: rowData.variable}">
-                    {{ col.textContent || rowData[col.fieldName] }}
+                    {{col.textContent}} {{rowData[col.fieldName]}}
```
fmontes: This could print out an extra space when `textContent` is not coming, you need to leave it as it was and just pass the `View X` in the `textContent` when you are mapping that.

PR #1247, opened 2019-11-25
issue add Query method to ApiContent dotcms lib

alfredo-dotcms:


projects/dotcms/src/lib/api/DotApiContent.ts
```diff
@@ -1,5 +1,22 @@
...
+    const attrs = {
+        depth: `/depth/${params.depth}`,
+        limit: `/limit/${params.limit}`,
+        offset: `/offset/${params.offset}`,
+        orderBy: `/orderby/${params.orderBy}`
+    };
+
+    for (const key of Object.keys(params)) {
+        url += key !== 'query' ? attrs[key] : '';
+    }
```
freddyucv: I thin is better do this like:

`${url}/depth/${params.depth}/limit/${params.limit}/offset/${params.offset}/orderby/${params.orderBy}`

projects/dotcms/src/lib/api/DotApiContent.ts
```diff
@@ -22,12 +45,12 @@ export class DotApiContent {
...
     ): Promise<Response> {
         const response = await this.dotCMSHttpClient.request({
             url,
-            method: 'POST',
-            body: JSON.stringify(params)
+            method: params ? 'POST' : 'GET',
+            body: params ? JSON.stringify(params) : ''
         });
 
         if (response.status !== 200) {
```
freddyucv: test?

projects/dotcms/src/lib/api/DotApiContent.ts
```diff
@@ -1,5 +1,22 @@
...
+        offset: `/offset/${params.offset}`,
+        orderBy: `/orderby/${params.orderBy}`
+    };
+
+    for (const key of Object.keys(params)) {
+        url += key !== 'query' ? attrs[key] : '';
+    }
+
+    return url;
+}
```
freddyucv: test?

projects/dotcms/src/lib/api/DotApiContent.ts
alfredo-dotcms: if not all params are received, then an Url with not needed params will be sent:
`${url}/depth/limit/offset/orderby/`

projects/dotcms/src/lib/api/DotApiContent.ts
```diff
@@ -22,12 +45,12 @@ export class DotApiContent {
...
     private async doRequest<Content extends DotCMSContent>(
         url: string,
-        params: Content
+        params?: Content
     ): Promise<Response> {
         const response = await this.dotCMSHttpClient.request({
             url,
-            method: 'POST',
-            body: JSON.stringify(params)
+            method: params ? 'POST' : 'GET',
```
fmontes: This is not a good approach, conceptually it doesn't make sense that because we sent or not `params` the method change.

projects/dotcms/src/lib/api/DotApiContent.ts
```diff
@@ -12,6 +29,12 @@ export class DotApiContent {
         this.dotCMSHttpClient = httpClient;
     }
 
+    query(params: DotCMSContentQuery): Promise<Response> {
```
fmontes: doc.

projects/dotcms/src/lib/models/DotCMSContent.model.ts
```diff
@@ -1,3 +1,11 @@
 export interface DotCMSContent {
     stName: string;
 }
+
+export interface DotCMSContentQuery {
```
fmontes:
```typescript
export interface DotCMSContentQuery {
    query: string;
    options: {
        depth?: string;
        limit?: string;
        offset?: string;
        orderBy?: string;
    }
}
```

If you do it this way then here you can do `populateQueryUrl(params.options)` and you don't have to do the logic of avoid the property `query`

projects/dotcms/src/lib/api/DotApiContent.ts
```diff
@@ -12,6 +29,12 @@ export class DotApiContent {
...
+        let url = `/api/content/query/${params.query}`;
```
fmontes:
```suggestion
        let url = `/api/content/query/${params.query}${populateQueryUrl(params)}`;
```

projects/dotcms/src/lib/api/DotApiContent.ts
freddyucv: But if a params is not receive,  an Url with not needed params will be sent with this code

PR #1249, opened 2019-12-02
dotCMS/core#17676 Refresh the portlet when bundles gets deleted

fmontes:
Listen for a specific websocket event and send a global "reload" event to the iframe user using this: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

### DIFF
--- a/src/app/view/components/_common/iframe/iframe-component/iframe.component.ts
+++ b/src/app/view/components/_common/iframe/iframe-component/iframe.component.ts
@@ -10,15 +10,16 @@ import {
     OnDestroy
 } from '@angular/core';
 
-import { takeUntil } from 'rxjs/operators';
+import { takeUntil, filter } from 'rxjs/operators';
 import { Subject } from 'rxjs';
 
-import { LoginService } from 'dotcms-js';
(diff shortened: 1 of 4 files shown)


src/app/view/components/_common/iframe/iframe-component/iframe.component.ts
```diff
@@ -111,6 +117,67 @@ export class IframeComponent implements OnInit, OnDestroy {
...
+            .subscribe((event: DotEventTypeWrapper<any>) => {
+                this.loggerService.debug('Capturing Site Browser event', event.name, event.data);
+            });
+
+        webSocketEvents$
+            .pipe(
+                filter(
+                    (event: DotEventTypeWrapper<any>) =>
+                        event.name === 'DELETE_BUNDLE' &&
+                        this.dotRouterService.currentPortlet.id === 'publishing-queue' &&
```
hmoreras: `this.dotRouterService.currentPortlet.id === 'publishing-queue'` this is not needed. since the refresh will be handled by the jsp

fmontes: True this! great catch.

PR #1250, opened 2019-12-02
Allow to receive language code or id, when getting a page. 

hmoreras:
We are assuming that `params.language` is always a code ( `en | es`), but can be cases where can be the language id directly. This fix is to take in consideration the second scenario.

### DIFF
--- a/projects/dotcms/src/lib/api/DotApiPage.ts
+++ b/projects/dotcms/src/lib/api/DotApiPage.ts
@@ -24,9 +24,9 @@ export class DotApiPage {
         if (params.language) {
             params = {
                 ...params,
-                language: isNaN(params.language as any) ? await this.dotAppLanguage.getId(params.language) : params.language
-                    ? params.language
-                    : await this.dotAppLanguage.getId(params.language)
+                language: isNaN(params.language as any)
+                    ? await this.dotAppLanguage.getId(params.language)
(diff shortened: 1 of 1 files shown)


projects/dotcms/src/lib/api/DotApiPage.ts
```diff
@@ -24,7 +24,9 @@ export class DotApiPage {
         if (params.language) {
             params = {
                 ...params,
-                language: await this.dotAppLanguage.getId(params.language)
+                language: !isNaN(params.language as any)
```
fmontes: Is not good approach to start a condition with the "deny" condition.

```suggestion
                language: isNaN(params.language as any) ? await this.dotAppLanguage.getId(params.language) : params.language
```

projects/dotcms/src/lib/api/DotApiPage.ts
```diff
@@ -24,7 +24,9 @@ export class DotApiPage {
         if (params.language) {
             params = {
                 ...params,
-                language: await this.dotAppLanguage.getId(params.language)
+                language: isNaN(params.language as any)
+                    ? await this.dotAppLanguage.getId(params.language)
+                    : params.language
             };
         }
```
freddyucv: test?

PR #1260, opened 2019-12-20
dotCMS/core#17722 Show submenu as flyout when menu is collapsed

hmoreras:


src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.scss
```diff
@@ -5,6 +5,29 @@
...
+            .dot-nav__tooltip {
+                display: none !important;
+            }
+        }
+    }
+
+    ::ng-deep {
+        .ui-tooltip {
+            &.dot-nav__tooltip {
+                margin-left: $basic-padding + $navigation-width-collased - $navigation-width;
```
fmontes: Isn't this dynamically set by the tooltip directive?

src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.scss
```diff
@@ -22,6 +51,48 @@
 
         dot-sub-nav {
             width: $navigation-width-collased;
+
+            ::ng-deep .dot-nav-sub {
```
fmontes: Why here and not in `dot-sub-nav.component.scss`?

src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.scss
```diff
@@ -22,6 +51,48 @@
...
+                background: $white;
+                box-shadow: $md-shadow-1;
+
+                ::ng-deep .dot-nav-sub {
+                    height: auto !important;
+                    overflow: visible !important;
+                }
+            }
+
+            dot-overlay-mask {
```
fmontes: Shouldn't this be same for the whole component?

hmoreras: need's to be removed

src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.scss
hmoreras: because of this class, https://github.com/dotCMS/core-web/pull/1260/files#diff-8a33fecd3be9bb9d34dfa2c8b84590efR31

src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.scss
hmoreras: no with the current implementation, because part of the target element is hidden. maybe we need to change the current approach ?

src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.scss
hmoreras: changed.

src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.html
```diff
@@ -14,4 +14,8 @@
         aria-hidden="true">
     </dot-icon>
 </div>
-<dot-sub-nav [data]="data" (itemClick)="handleItemClick($event)"></dot-sub-nav>
+<dot-sub-nav [data]="data"
+             [class.collapsed]="collapsed"
+             [class.contextmenu]="contextmenu"
```
fmontes: Is simpler to pass the `collapsed` as an `@Input` to `dot-sub-nav` and handle all the styling INSIDE the `dot-sub-nav` which is where it supposed.

Basically the idea is to tell the `dot-sub-nav` that works differently if `collapsed=true`.

src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.ts
```diff
@@ -28,5 +57,55 @@ export class DotNavItemComponent {
             originalEvent: $event,
             data: data
         });
+        this.dotEventsService.notify('hide-sub-nav-fly-out');
```
freddyucv: could 'hide-sub-nav-fly-out' be a const?

src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.ts
```diff
@@ -28,5 +57,55 @@ export class DotNavItemComponent {
...
+     *
+     * @param { originalEvent: MouseEvent; data: DotMenuItem } $event
+     * @memberof DotNavItemComponent
+     */
+    handleItemClick(event: { originalEvent: MouseEvent; data: DotMenuItem }) {
+        this.itemClick.emit(event);
+        this.dotEventsService.notify('hide-sub-nav-fly-out');
+    }
+
+    private setHideFlyOutSubscription(): void {
```
freddyucv: I think this method's name not should have the set prefix because I expect that a method with set... receive a params to be set

PR #1268, opened 2020-01-17
dotCMS/core#17828 Chrome Site selector is not refreshing 

hmoreras:


projects/dotcms-js/src/lib/core/site.service.ts
```diff
@@ -147,13 +148,11 @@ export class SiteService {
      */
     switchSite(site: Site): void {
         this.loggerService.debug('Applying a Site Switch', site.identifier);
-
         this.coreWebService
             .requestView({
                 method: RequestMethod.Put,
                 url: `${this.urls.switchSiteUrl}/${site.identifier}`
-            })
-            .subscribe(() => this.setCurrentSite(site));
```
fmontes: Don't we need subscribe here?

hmoreras: no, because we now hear the `SWITCH_SITE ` event. https://github.com/dotCMS/core-web/pull/1268/files#diff-7a9328737943182522e1d2e3d7165d0dR52

projects/dotcms-js/src/lib/core/site.service.ts
```diff
@@ -178,11 +177,10 @@ export class SiteService {
     }
 
     private setCurrentSite(site: Site): void {
+        this.selectedSite = site;
         if (this.selectedSite) {
```
alfredo-dotcms: is the `IF` here makes sense, now?

you're assigning it on the previous line

hmoreras: true, I did this  change because we where updating the `_switchSite` while the `selectedSite ` was not updated yet.

PR #1304, opened 2020-04-21
feat: allow custom fetch for dotcms requests

samuelmaddock:
These changes add support for using a custom fetch implementation and adds `node-fetch` types.

The use case I have is that our dotCMS instance is hosted behind a load balancer. Each request to the dotCMS API will switch nodes round-robin unless cookies are stored between requests. I'd like to use [`fetch-cookie`](https://www.npmjs.com/package/fetch-cookie) with `node-fetch` to store cookies, thus sticking with one node instead of alternating between requests.

Usage:
```ts
const dotCmsApi = initDotCMS({
  host: process.env.DOTCMS_HOST,
  token: process.env.DOTCMS_TOKEN,
  fetch: customFetch
})
```

### DIFF
--- a/projects/dotcms/src/lib/api/DotApiConfiguration.ts
+++ b/projects/dotcms/src/lib/api/DotApiConfiguration.ts
@@ -18,7 +18,7 @@ export class DotApiConfiguration {
             .request({
                 url: '/api/v1/configuration'
             })
-            .then(async (response: Response) => {
+            .then(async (response) => {
                 if (response.status === 200) {
                     const data = await response.json();
                     return data.entity;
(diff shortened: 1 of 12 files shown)


projects/dotcms/src/lib/api/DotApiConfiguration.ts
```diff
@@ -18,7 +18,7 @@ export class DotApiConfiguration {
             .request({
                 url: '/api/v1/configuration'
             })
-            .then(async (response: Response) => {
```
freddyucv, line 21: why do you remove the Response type?

samuelmaddock: Previously `dotCMSHttpClient.request` returned `Promise<any>` which required adding the `Response` type here. Now that `@types/node-fetch` is included, these types can be inferred.

fmontes:
https://github.com/dotCMS/core/issues/20159

PR #1321, opened 2020-05-06
Fix Tests in Master. 

hmoreras:
Just fixing tests failing in travis.


src/app/view/components/_common/dot-alert-confirm/dot-alert-confirm.spec.ts
```diff
@@ -66,7 +66,6 @@ describe('DotAlertConfirmComponent', () => {
 
             fixture.detectChanges();
             const confirm = de.query(By.css('p-confirmDialog')).componentInstance;
-            expect(confirm.responsive).toBe(true, 'responsive');
```
hmoreras, line 69: `p-confirmDialog` do not support `responsive` property anymore, the attr was removed at some point in the html

PR #1333, opened 2020-05-20
dotCMS/core#17921 fix Center closed nav icons

alfredo-dotcms:


src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.scss
```diff
@@ -69,7 +69,7 @@
     display: flex;
     justify-content: space-between;
     min-height: $navigation-item-height;
-    padding: 14px $basic-padding * 3 13px 23px; // top and bottom padding to get min 48px height
+    padding: 14px 22px 13px 25px; // top and bottom padding to get min 48px height
```
fmontes, line 72: We can't hardcode this to pixels because here you can have FA icons but also MD so we need to find a way to make both to be centered.

alfredo-dotcms:
@fmontes when using FA or MD icons, the container of the icon `dot-nav-icon` size will always be 18x18, no matter what icon it has inside it will centered, so using a hardcoded pixels in the padding is fine

On a new commit I added a size = 18, that IMO was missing, in that way no matter what type of icon you choose it will be set to size 18 and be consistent (before MD was size 18 and FA was size 14)

image

PR #1348, opened 2020-06-12
issue-18642-create-select-field-dotApps

alfredo-dotcms:


src/app/shared/models/dot-apps/dot-apps.model.ts
```diff
@@ -35,3 +36,8 @@ export interface DotAppsSaveData {
         value: string;
     };
 }
+
+interface LabelValueOption {
```
hmoreras, line 40: We use the interface `SelectItem` from primeNg in most components. Should we changed ?

PR #1355, opened 2020-06-17
dotCMS/core#18531 Create new message service

hmoreras:
always request for messages on login. 
add tests to dot-message.service


src/app/view/components/login/dot-login-component/dot-login.component.ts
```diff
@@ -40,6 +40,7 @@ export class DotLoginComponent implements OnInit, OnDestroy {
     ) {}
 
     ngOnInit() {
+        this.dotMessageService.init('default');
```
fmontes, line 43: Why don't we set the default inside the service.

hmoreras: is does by default, without params, but when we send the parameter we "force" to make a request and refresh the local storage

PR #1361, opened 2020-06-19
dotCMS/core#18716 5.3.2 Main nav does not scoll

fmontes:
In the collapsed version what I'm doing is checking the window height and the submenu position and base on that align it the sub menu.

In the expend version is just on line of css.

### DIFF
--- a/src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.ts
+++ b/src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.ts
@@ -46,6 +46,7 @@ export class DotNavItemComponent {
      * @memberof DotNavItemComponent
      */
     setSubMenuPosition(): void {
+
         if (this.collapsed) {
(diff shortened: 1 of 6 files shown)


src/app/view/components/dot-navigation/components/dot-nav-item/dot-nav-item.component.ts
```diff
@@ -35,6 +40,27 @@ export class DotNavItemComponent {
...
+     *
+     * @memberof DotNavItemComponent
+     */
+    setSubMenuPosition(): void {
+
+        if (this.collapsed) {
+            const [rects] = this.subnav.ul.nativeElement.getClientRects();
+
+            if (rects.bottom > window.innerHeight) {
+                this.customStyles = {
```
alfredo-dotcms, line 54: why not moving this to a CSS class and then based on the IF you add the css class or not

fmontes: Because I want to make sure is not overwritten by any other class.

PR #1368, opened 2020-07-01
#18714 Fixing error when try to add the same file container twice int…

freddyucv:
When a layout had a sidebar end this sidebar has containers into in, then to the value was taking the id always no matter the containers type

https://github.com/dotCMS/core-web/compare/issue-18714-Container-id-is-being-duplicated?expand=1#diff-2137771d821e30c8a489a7cceb39301eR44

Also we was not allowing add twice the same container into the sidebar

https://github.com/dotCMS/core-web/compare/issue-18714-Container-id-is-being-duplicated?expand=1#diff-fb095d9ba9f6d24e2d155df72d98236fR3


dsilvam:
dotCMS/core#18174

PR #1371, opened 2020-07-08
dotCMS/core#18779 Apps - UI: Show the Enterprise Screen when the license level is Community

alfredo-dotcms:


src/app/portlets/dot-apps/dot-apps-list/dot-apps-list.component.ts
```diff
@@ -58,6 +53,29 @@ export class DotAppsListComponent implements OnInit, OnDestroy {
...
+    private getApps(): void {
+        this.dotAppsService
+            .get()
+            .pipe(takeUntil(this.destroy$))
+            .subscribe((apps: DotApps[]) => {
+                this.apps = apps;
+                this.appsCopy = _.cloneDeep(apps);
+                setTimeout(() => {
+                    this.attachFilterEvents();
+                }, 0);
```
freddyucv, line 65: why do we need this timeout?

alfredo-dotcms: since `ng-template` is being used in the HTML, the binding of events needs to wait for the DOM to be rendered, otherwise when attaching the `KeyUp` and `focus` to the **searchInput** will break

PR #1392, opened 2020-07-31
dots are not centered in action button in edit mode

alfredo-dotcms:


src/styles/primeng-theme/components/_button.scss
```diff
@@ -210,7 +210,6 @@
 .ui-button-icon-only {
     .fa {
         margin-top: -0.5em;
-        margin-left: -0.65em;
```
fmontes, line 213: Isn't a way to do this without a set number?

alfredo-dotcms: actually removed all fixed margins, and then compared on the app and it does not affects any icons

PR #1413, opened 2020-08-18
dotCMS/core#19117 Unable to open the task details on the task portlet

fmontes:


src/app/view/components/dot-workflow-task-detail/services/dot-workflow-task-detail.service.ts
```diff
@@ -16,8 +16,8 @@ export interface DotTaskAction {
  */
 @Injectable()
 export class DotWorkflowTaskDetailService {
-    private data: Subject<DotTaskAction> = new Subject();
-    private _header: Subject<string> = new Subject();
+    private data: BehaviorSubject<DotTaskAction> = new BehaviorSubject(null);
```
fmontes, line 19: This is the fix, there was a race condition so when we subscribe to the `viewUrl$` the Observable was already sent, by using `BehaviorSubject` we prevent that because it will always sent you the current value on subscription.

PR #1426, opened 2020-09-02
dotCMS/core#19220 #editContentlet fail in edit mode when added it directly to the advanced template

fmontes:
The fix here is that if we have the `#editContentlet` outside a contentlet we can reload just that contentlet after edit so we have to reload the whole page.


hmoreras (approved):

src/app/portlets/dot-edit-page/content/services/dot-edit-content-html/dot-edit-content-html.service.ts
```diff
@@ -69,6 +69,7 @@ export class DotEditContentHtmlService {
                     | DotContentletEventSelect
                     | DotContentletEventSave
             ) => {
+                console.log('here', this)
```
alfredo-dotcms, line 72: remove

alfredo-dotcms (approved):

PR #1436, opened 2020-09-21
#19168 Added GitHub Actions and script to push to test-results repo

ivorpad:
Added a GitHub Action to run tests in GitHub instead of Travis


package.json
```diff
@@ -7,6 +7,7 @@
         "lint": "ng lint",
         "e2e": "ng e2e",
         "test": "ng test",
+        "test-headless": "ng test dotcms-ui --watch=false --browsers=ChromeHeadless",
```
fmontes, line 10: why we used to run in travis?

.cicd/github.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
fmontes, line 1: We need to remove the travis so we don't run the tests in travis too.

fmontes (changes requested):

PR #1473, opened 2020-10-19
Fix data table in content types

fmontes:


src/app/portlets/shared/dot-content-types-edit/components/fields/content-type-fields-properties-form/content-type-fields-properties-form.component.scss
```diff
@@ -2,7 +2,7 @@
 
 :host {
     display: block;
-    width: 500px;
+    width: 600px;
```
ivorpad, line 5: Why not change `px` for `rems` from now on?

src/app/view/components/dot-listing-data-table/dot-listing-data-table.component.scss
```diff
@@ -40,22 +40,24 @@
...
-        }
+.listing-datatable__column-icon {
+    display: flex;
+    align-items: center;
 
-        dot-action-header {
-            background: $white;
-        }
+    dot-icon {
+        margin-right: 8px;
```
ivorpad, line 51: `$spacing-1` is `.5rem` or `7px`. Do we really need `8px` here?

src/styles/dotcms-theme/components/_datatable.scss
```diff
@@ -69,12 +78,12 @@
 }
 
 .p-datatable .p-sortable-column:not(.p-highlight):hover {
-    background: $hover-bg;
+    background: rgba(var(--color-sec_rgb), 0.05);
```
ivorpad, line 81: `$brand-secondary_rgb` instead of `--color-sec_rgb` ? Why not move this to a SCSS variable? I'm seeing below more `$brand-secondary_rgb, 0.05`

src/app/portlets/shared/dot-content-types-edit/components/fields/content-type-fields-properties-form/content-type-fields-properties-form.component.scss
fmontes: True this, fotgot about it.

src/styles/dotcms-theme/components/_datatable.scss
fmontes: Not sure how to handle this one because is like the hover and focus state just for the table and I don't see us using it anywhere else... do you think it would be good to have the var inline in this same file?

ivorpad: What if we do:

```scss
@use 'variables' as * with (
  $hover-bg: rgba($brand-secondary_rgb, 0.05),
);
```

Not sure if `as * with` works though, but it's worth a try.

PR #1474, opened 2020-10-19
fix dot page selector - remove MdInputTextModule

hmoreras:
fit dot-page-selector, remove MdInputTextModule.


src/styles/_mixins.scss
```diff
@@ -100,7 +100,7 @@
 @mixin dot-field-helper-on-input {
     position: absolute;
     right: 0;
-    top: 19px;
+    top: $basic-padding /2;
```
fmontes, line 103: Try this better

```
.parent {
  position: relative;
}
.child {
  position: absolute;
  top: 50%;
  transform: translateY(-50%);
}
```

Because if we change the height of the input we don't have to come here and fix the top.

fmontes (changes requested):

PR #1484, opened 2020-10-21
Replace basic-padding with spacings

fmontes:


projects/dot-rules/src/lib/styles/rule-engine.scss
```diff
@@ -35,6 +33,9 @@ $brand-primary_mod: var(--color-main_mod);
 $md-shadow-1: 0 1px 3px rgba(0, 0, 0, 0.12), 0 1px 2px rgba(0, 0, 0, 0.24);
 $md-shadow-3: 0 10px 24px 0 rgba(0, 0, 0, 0.2);
 
+$spacing-3: 1rem;
```
fmontes, line 36: Needs to find a way reuse global variables here.

hmoreras (approved):

PR #1538, opened 2020-12-14
issue-19597-add-route-to-create-new-content-type

alfredo-dotcms:


src/app/app-routing.module.ts
```diff
@@ -104,6 +106,18 @@ const PORTLETS_IFRAME = [
                                 (m) => m.DotPortletDetailModule
                             ),
                         path: ':asset'
+                    },
+                    {
+                        path: 'new',
+                        children: [
+                            {
+                                component: DotCreateContentletComponent,
```
fmontes, line 114: This should be a lazy load module.

fmontes (changes requested):

src/app/api/services/dot-router/dot-router.service.spec.ts
```diff
@@ -96,6 +96,20 @@ describe('DotRouterService', () => {
...
+        expect(service.goToStarter).toHaveBeenCalledTimes(1);
+    });
+
+    it('should go to Content page', () => {
+        spyOn(service, 'goToContent');
+        service.goToContent();
+
+        expect(service.goToContent).toHaveBeenCalledTimes(1);
+    });
+
```
fmontes, line 112: This test is not right you are only testing that you called the method but not what happen inside the method, what you need to assert here is that when you call those methods the `this.router.navigate` was called with the specific url like this one: https://github.com/dotCMS/core-web/blob/9b20c831fda4c8d10e9083b3001f6325afce5fd5/src/app/api/services/dot-router/dot-router.service.spec.ts#L113

fmontes (changes requested):

hmoreras (approved):

fmontes (approved):

PR #1545, opened 2021-01-11
dotCMS/core#19668 Add history tab 

hmoreras:
Add history to history tab and support delete template and edit template from inode

### DIFF
--- a/src/app/api/services/dot-router/dot-router.service.ts
+++ b/src/app/api/services/dot-router/dot-router.service.ts
@@ -108,7 +108,8 @@ export class DotRouterService {
     }
 
     /**
-     * Redirect to edit the template id passed
+     * Redirect to edit the template.
+     * If the inode is paased, load a specific version of the template
      *
      * @param {string} id
      * @memberof DotRouterService
--- a/src/app/api/services/dot-router/dot-router.service.ts
+++ b/src/app/api/services/dot-router/dot-router.service.ts
@@ -114,7 +114,9 @@ export class DotRouterService {
      * @memberof DotRouterService
      */
     goToEditTemplate(id: string, inode?: string): void {
-        this.router.navigate([inode ? `/templates/edit/${id}/${inode}` : `/templates/edit/${id}`]);
+        this.router.navigate([
+            inode ? `/templates/edit/${id}/inode/${inode}` : `/templates/edit/${id}`
+        ]);
     }
 
     goToURL(url: string): void {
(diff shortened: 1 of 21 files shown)


ivorpad (approved):

src/app/api/services/dot-router/dot-router.service.ts
```diff
@@ -113,8 +113,8 @@ export class DotRouterService {
      * @param {string} id
      * @memberof DotRouterService
      */
-    goToEditTemplate(id: string): void {
-        this.router.navigate([`/templates/edit/${id}`]);
+    goToEditTemplate(id: string, inode?: string): void {
```
fmontes, line 116: update doc.

src/app/api/services/dot-templates/dot-templates.service.ts
```diff
@@ -47,6 +47,20 @@ export class DotTemplatesService {
         });
     }
 
+    /**
+     * Get the template, tittle or inode .
```
fmontes, line 51: Not sure that desc explain what this is doing.

src/app/portlets/dot-templates/dot-template-create-edit/dot-template-builder/dot-template-builder.component.ts
```diff
@@ -13,10 +14,20 @@ export class DotTemplateBuilderComponent implements OnInit {
     permissionsUrl = '';
     historyUrl = '';
 
-    constructor() {}
+    constructor(private dotCustomEventHandlerService: DotCustomEventHandlerService) {}
```
fmontes, line 17: This is a dumb component, let's keep it that way, let's just emit the custom event out so `DotTemplateCreateEditComponent` which is the parent does the redirect to edit the template.

Also we don't need to modify `DotCustomEventHandlerService` let's keep all centralize into the templates portlet just do the redirect using the router service in `DotTemplateCreateEditComponent`.

src/app/portlets/dot-templates/dot-templates-routing.module.ts
```diff
@@ -28,6 +28,16 @@ const routes: Routes = [
         resolve: {
             template: DotTemplateCreateEditResolver
         }
+    },
+    {
+        path: 'edit/:id/:inode',
```
fmontes, line 33: I think the path should be `edit/:id?inode="xxxx"`

Because like this is too confusing for the user, using a queryparam we indicate the user we are passing an `inode`.

And also using `queryparam` we don't need to create a new route.

fmontes (changes requested):

PR #1602, opened 2021-03-01
dotCMS/core#20022 Fix JS error while creating a template.

hmoreras:
check `this.paginatorService.url` is set to avoid the first event of `pageChange` when initially loads, then the `this.paginatorService.url` is set on the `show` event.

### DIFF
--- a/projects/dotcms-js/src/lib/core/core-web.service.ts
+++ b/projects/dotcms-js/src/lib/core/core-web.service.ts
@@ -261,15 +261,14 @@ export class CoreWebService {
     }
 
     private getFixedUrl(url: string): string {
-        if (url.startsWith('api')) {
+        if (url?.startsWith('api')) {
             return `/${url}`;
         }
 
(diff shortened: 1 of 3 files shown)


projects/dotcms-js/src/lib/core/core-web.service.ts
```diff
@@ -261,15 +261,14 @@ export class CoreWebService {
     }
 
     private getFixedUrl(url: string): string {
-        if (url.startsWith('api')) {
+        if (url?.startsWith('api')) {
             return `/${url}`;
         }
 
-        const [version] = url.split('/');
+        const version = url ? url.split('/')[0] : '';
```
hmoreras, line 268: this way because using just `?` + `[version]`, gives, https://stackoverflow.com/questions/50329128/what-is-an-intermediate-value

fmontes (approved):

alfredo-dotcms (approved):

PR #1730, opened 2021-08-13
dotCMS/core#20788 Listing row should have the same min-height

rjvelazco:
### Proposed Changes
* change 1
* change 2

### Checklist
- [ ] Tests
- [ ] Translations

### Additional Info

### Screenshots
 |

### DIFF
--- a/libs/dot-primeng-theme-styles/src/scss/dotcms-theme/components/_datatable.scss
+++ b/libs/dot-primeng-theme-styles/src/scss/dotcms-theme/components/_datatable.scss
@@ -100,6 +100,7 @@ $datatable-bg-hover: rgba($brand-secondary_rgb, 0.05);
 .p-datatable .p-datatable-tbody > tr {
     background: $white;
     color: $black;
+    height: 4rem;
     transition: none;
     outline-color: transparent;
     cursor: pointer;


libs/dot-primeng-theme-styles/src/scss/dotcms-theme/components/_datatable.scss
```diff
@@ -100,6 +100,7 @@ $datatable-bg-hover: rgba($brand-secondary_rgb, 0.05);
 .p-datatable .p-datatable-tbody > tr {
     background: $white;
     color: $black;
+    height: 4rem;
```
fmontes, line 103: We can have height because if we have to have 2 lines it will break.

fmontes (changes requested):

libs/dot-primeng-theme-styles/src/scss/dotcms-theme/components/_datatable.scss
rjvelazco: The reason why I used `height` instead of `min-height` is because that a table cell will always stretch if necessary, so height actually is the equivalent of min-height for a table cell.

PR #1775, opened 2021-10-11
dotCMS/core#21014 fix Combined Default Content View in Content Search…

alfredo-dotcms:
The customer would like to add a default view that combines these two, showing both a thumbnail preview and contentlet data, returning something like the mock up below:

image

### Checklist
- [ ] Tests
- [ ] Translations

### Additional Info

### Screenshots
** original screenshot **  |  ** updated screenshot **

### DIFF
--- a/libs/dotcms-webcomponents/src/elements/dot-contentlet-icon/dot-contentlet-icon.tsx
+++ b/libs/dotcms-webcomponents/src/elements/dot-contentlet-icon/dot-contentlet-icon.tsx
@@ -128,7 +128,10 @@ export class DotContentletIcon {
     private ext: string;
 
     componentWillRender() {
-        // if doesn't match "Icon", then the Material Icon name is passed and it should be displayed as is
+        /*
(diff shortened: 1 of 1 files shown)


libs/dotcms-webcomponents/src/elements/dot-contentlet-icon/dot-contentlet-icon.tsx
```diff
@@ -128,15 +128,17 @@ export class DotContentletIcon {
     private ext: string;
 
     componentWillRender() {
-        this.ext = this.icon.replace('Icon', '');
+        // if doesn't match "Icon", then the Material Icon name is passed and it should be displayed as is
```
fmontes, line 131:
```suggestion
        /*
        If it doesn't contain "Icon" (with uppercase) we're assuming that is coming to a material icon name, which is the only way we have to differentiate between new and legacy icons without passing an extra attribute.
        */
```

fmontes (changes requested):

PR #1798, opened 2021-11-18
dotCMS/core#21259 Add "Link" button to bubble menu 

rjvelazco:


libs/block-editor/src/lib/demo/block-editor.component.ts
```diff
@@ -41,16 +47,20 @@ export class BlockEditorComponent implements OnInit {
...
+                TextAlign.configure({ types: ['heading', 'paragraph', 'listItem'] }),
+                Highlight.configure({ HTMLAttributes: { class: 'highlighted'} }),
+                Link.configure({ openOnClick: true })
             ]
         });
     }
+
+    openForm() {
+        this.linkForm.showForm();
+    }
```
fmontes, line 65: Like my previous comment, let's load this dynamically, and hopefully, we can have a nice fade-in animation.

libs/block-editor/src/lib/demo/block-editor.component.scss
```diff
@@ -12,6 +12,11 @@ tiptap-editor {
         display: block;
         height: 100%;
         border-radius: 4px;
-        padding: 1rem 4rem;
+        padding: 1.5rem 4rem;
     }
 }
+
+.highlighted {
```
fmontes, line 19: What is this class for?

libs/block-editor/src/lib/demo/block-editor.component.html
```diff
@@ -1,4 +1,5 @@
 <tiptap-bubble-menu [editor]="editor">
```
fmontes, line 1: Little bug:

https://user-images.githubusercontent.com/751424/142462537-65f1930e-c43b-4d4d-a6b5-7d5b3a5b0907.mp4

fmontes: Tippy showing twice.

https://user-images.githubusercontent.com/751424/142465220-1dd06b15-5c24-4277-846c-58278c50562d.mp4

libs/block-editor/src/lib/extensions/components/bubble-menu-link-form/bubble-menu-link-form.component.html
```diff
@@ -0,0 +1,50 @@
...
+    (click)="hideForm()"
+    #formContainer
+>
+    <form
+        autoComplete="off"
+        (click)="preventHideOnFormClick( $event )"
+        (mousedown)="focusInput()"
+        (submit)="addLink()"
+        #form
+    >
```
fmontes, line 13: You don't need to nest the form inside the overlay that is what causes that you have to do `preventHideOnFormClick`.

But now I think, shouldn't we listen to a click or a focus in the editor to hide our things? instead of creating a div just for this?

libs/block-editor/src/lib/extensions/components/bubble-menu-link-form/bubble-menu-link-form.component.html
```diff
@@ -0,0 +1,50 @@
...
+            />
+        </div>
+        <div *ngIf="nodeLink">
+            <div class="info-box">
+                <span class="sub-title">LINKED TO</span>
+                <div class="link-box">
+                    <a [href]="nodeLink">
+                        <span class="material-icons">language</span>
+                        <span>
+                            {{ nodeLink.length > 35 ? (nodeLink | slice:0:35 ) + '...' : nodeLink}}
```
fmontes, line 32: TODAY I LEARN, there is a `slice` pipe 😬

libs/block-editor/src/lib/extensions/components/bubble-menu-link-form/bubble-menu-link-form.component.scss
```diff
@@ -0,0 +1,95 @@
...
+            padding: 10px;
+            width: 100%;
+            span {
+                font-size: 1.25rem;
+            }
+            &:hover {
+                background-color: #f8f9fa;
+            }
+        }
+    }
```
fmontes, line 94: Doesn't need to be nested, Angular will encapsulate it automagically.

libs/block-editor/src/lib/extensions/components/bubble-menu-link-form/bubble-menu-link-form.component.ts
```diff
@@ -0,0 +1,119 @@
...
+            duration: 150,
+            getReferenceClientRect: null,
+            content: this.form.nativeElement,
+            interactive: true,
+            trigger: 'manual',
+            placement: 'bottom',
+            hideOnClick: 'toggle',
+            animation: 'fade',
+            appendTo: this.formContainer.nativeElement
+        });
```
fmontes, line 46: For what I understand, tippy should be at the editor level, we just pass to the tippy element, we should not be passing editor to this component, this should be a dumb component.

We should just emit events and at the extension, level do our things.

fmontes (changes requested):

libs/block-editor/src/lib/demo/block-editor.component.scss
rjvelazco: This class is for the highlight of the text to which the link is going to be added.

When you click on the form to add a link, we lost the text selection. That class is added to the text to simulate a text selection.

Screen Shot 2021-11-18 at 2 26 13 PM

libs/block-editor/src/lib/extensions/components/bubble-menu-link-form/bubble-menu-link-form.component.html
rjvelazco: The purpose of a div is to avoid clicking something else when the `add-link-form` is ope

https://user-images.githubusercontent.com/72418962/142477447-985a0eae-27d9-484d-9e49-a04c3922faaa.mov

n.

libs/block-editor/src/lib/demo/block-editor.component.html
rjvelazco: Mmm... I'll take a look at that.

libs/block-editor/src/lib/extensions/components/bubble-menu-link-form/bubble-menu-link-form.component.html
fmontes: Exactly, the `div` is to have an area to click and close the form but because you put the form INSIDE this div then clicks in the form will trigger clicks in the div and by that means the form close and in order to prevent that you had to add a click to the form to stoppropagation, instead of this you can just move the form outside the div and you will not have to care for that propagation.

But also, do we need the div or can we just listen for click in the editor itself?

libs/block-editor/src/lib/demo/block-editor.component.html
fmontes: Also, we need to make that when the selected text is an URL we pre-populate the link input with that url.

libs/block-editor/src/lib/demo/block-editor.component.ts
```diff
@@ -42,21 +56,116 @@ export class BlockEditorComponent implements OnInit {
...
+
+    focusInputLink() {
+        this.componentLinkForm.instance.focusInput();
+    }
+
+    private getNodeLink(): string {
+        return this.editor.isActive('link')
+            ? this.editor.getAttributes('link').href
+            : '';
```
fmontes, line 169: All this code should not be here, this should be an extension or something, not in the block editor, we need to check this.

fmontes (changes requested):

libs/block-editor/src/lib/plugins/bubble-link-form.plugin.ts
```diff
@@ -0,0 +1,233 @@
...
+    tippyOptions?: Partial<Props>;
+    storage: PluginStorage;
+    component?: ComponentRef<BubbleMenuLinkFormComponent>;
+}
+
+export type BubbleLinkFormViewProps = BubbleLinkFormProps & {
+    view: EditorView;
+};
+
+export class bubbleLinkFormView {
```
fmontes, line 29:
```suggestion
export class BubbleLinkFormView {
```

Classes name should be capital-case.

fmontes (changes requested):

fmontes (dismissed):

libs/block-editor/src/lib/extensions/components/bubble-menu-link-form/bubble-menu-link-form.component.scss
```diff
@@ -0,0 +1,106 @@
+@use "libs/dot-primeng-theme-styles/src/scss/variables" as *;
+
+.form-container {
+    background: $white;
+    border-radius: $border-radius;
+    box-shadow: 0px 4px 10px rgba(10, 7, 37, 0.1);
+    display: flex;
+    flex-direction: column;
+    padding: 0.5rem;
```
hmoreras, line 9: spacing  var in all tis file.

libs/block-editor/src/lib/plugins/bubble-link-form.plugin.ts
```diff
@@ -0,0 +1,233 @@
...
+        this.createTooltip();
+    }
+
+    update(view: EditorView, prevState?: EditorState): void {
+        const prePluginState = this.pluginKey.getState(view.state);
+        const currentPluginState = this.pluginKey.getState(prevState);
+
+        // Check that the plugin state is different
+        if (prePluginState.toggle != currentPluginState.toggle) {
+            if (this.tippy.state.isVisible) {
```
hmoreras, line 79: ternary operator here

hmoreras (changes requested):

hmoreras (dismissed):

PR #1817, opened 2021-12-14
dotCMS/core#21276 Suggestions menu with many options is broken

rjvelazco:


libs/block-editor/src/lib/extensions/components/suggestions-list-item/suggestions-list-item.component.ts
```diff
@@ -37,4 +37,8 @@ export class SuggestionsListItemComponent implements FocusableOption, OnInit {
     unfocus() {
         this.element.nativeElement.style = '';
     }
+
+    scrollIntoView() {
+        this.element.nativeElement.scrollIntoViewIfNeeded(false);
```
fmontes, line 42: We need to support FF: https://caniuse.com/?search=scrollIntoViewIfNeeded

libs/block-editor/src/lib/extensions/components/suggestions-list-item/suggestions-list-item.component.scss
```diff
@@ -1,10 +1,15 @@
 :host {
+    box-sizing: border-box;
     display: flex;
     gap: 1rem;
     padding: 0.25rem 0.5rem;
     background-color: #fff;
     cursor: pointer;
     align-items: center;
+    height: 40px;
```
fmontes, line 9: Why do we need this?

libs/block-editor/src/lib/extensions/components/suggestions-list-item/suggestions-list-item.component.scss
```diff
@@ -1,10 +1,15 @@
...
+    &:hover {
+        background: #eee;
```
fmontes, line 11: var?

fmontes (changes requested):

alfredo-dotcms (dismissed):

libs/block-editor/src/lib/extensions/components/suggestions-list-item/suggestions-list-item.component.ts
rjvelazco: I changed `scrollIntoViewIfNeeded` for `scrollIntoView` which is supported in FF (https://caniuse.com/?search=scrollIntoView)

libs/block-editor/src/lib/extensions/components/suggestions-list-item/suggestions-list-item.component.scss
rjvelazco: Suggestions-list-item had a non-explicit `height` of `39.Xpx`, I rounded it to 40px per list item to set a `max-height` of `400px` on the suggestion-list container to be able to display up to 10 options and more than that must be scroll.

PR #1866, opened 2022-02-01
dotCMS/core#19837 Fix lint - Step 1

fmontes:
### Proposed Changes
* Fix lint errors that are not manual

### Checklist
- [ ] Tests
- [ ] Translations

### Additional Info

### Screenshots
** original screenshot **  |  ** updated screenshot **


.eslintrc.json
```diff
@@ -9,6 +9,7 @@
                 "@nrwl/nx/enforce-module-boundaries": [
                     "error",
                     {
+                        "allowCircularSelfDependency": true,
```
fmontes, line 12: Nx recommend not to use mapped paths in the apps and only use relative paths but our `dotcms-ui` app is so big that we can use `../../../../../../`

PR #1923, opened 2022-03-15
dotCMS/core#21483 Update contentlet suggestions result

hmoreras:
### Proposed Changes
* Update UI of  contentlet suggestion: https://github.com/dotcms/core/issues/21483


libs/block-editor/src/lib/extensions/components/suggestions/suggestions.component.ts
```diff
@@ -28,6 +28,7 @@ export interface DotMenuItem extends Omit<MenuItem, 'icon'> {
     icon: string | SafeUrl;
     isActive?: () => boolean;
     attributes?: Record<string, unknown>;
+    data?: Record<string, unknown>;
```
rjvelazco, line 31: Why `Record<string, unknown>;`? it will not be a just `contentlet`? Or does it receive more data?

rjvelazco (changes requested):

rjvelazco (approved):

PR #1930, opened 2022-03-23
dotCMS/core#21884 Add margin between contentlet and image blocks 

hmoreras:
Result: 

image


rjvelazco (approved):

libs/block-editor/src/lib/extensions/blocks/contentlet-block/contentlet-block.component.scss
```diff
@@ -5,6 +5,7 @@ dotcms-contentlet-block {
     height: 100%;
     width: 100%;
     box-sizing: border-box;
+    margin-bottom: $spacing-1;
```
fmontes, line 8: Let's make it bigger... at least 1rem.

fmontes (changes requested):

PR #1933, opened 2022-03-28
fix: contentlet buttons doesn't show on edit mode with an EMA

rjvelazco:


fmontes:
This was a patch to 5.3.8.X

PR #1948, opened 2022-04-11
Issue 21695 problems pushing entire sites

freddyucv:
### Proposed Changes
Theme is not going to be required anymore for the Template


oidacra (approved):

apps/dotcms-ui/src/app/portlets/dot-templates/dot-template-create-edit/dot-template-props/dot-template-props.component.ts
```diff
@@ -28,7 +28,7 @@ export class DotTemplatePropsComponent implements OnInit {
                 ? {
                       ...template,
                       title: [template.title, Validators.required],
-                      theme: [template.theme, Validators.required]
```
fmontes, line 31: Can we have template with no themes now?

fmontes (changes requested):

fmontes:
@freddyucv can you elaborate on why themes are no longer required? Do we set a theme under the hood?

freddyucv:
> @freddyucv can you elaborate on why themes are no longer required? Do we set a theme under the hood? And why this affect pushing sites?

"can you elaborate on why themes are no longer required? Do we set a theme under the hood?", We want to create the template without set the theme, so you can set the theme later, if you use a template without theme into a page then we are going to take the system theme to render, you can see that in the backend change: https://github.com/dotCMS/core/pull/22005

" why this affect pushing sites?", the bug is happening when you try to PP the demo.dotcms.com site, this site has several template that have theme that does not exists, so when it is tried to be save in the receiver we got the error. 

Several of this  template are not been using at all I created a issue for that

https://github.com/dotCMS/core/issues/21257

PR #1959, opened 2022-04-21
dotCMS/core#22016 Allow keyboard filtering while adding blocks 

hmoreras:
### Proposed Changes
* Allow filtering after  the user type `/` in a new line.

### DIFF
--- a/libs/block-editor/src/lib/extensions/actions-menu.extension.ts
+++ b/libs/block-editor/src/lib/extensions/actions-menu.extension.ts
@@ -115,6 +115,7 @@ export const ActionsMenu = (viewContainerRef: ViewContainerRef) => {
     function onStart({ editor, range, clientRect }: SuggestionProps | FloatingActionsProps): void {
         suggestionsComponent = getSuggestionComponent(viewContainerRef);
         suggestionsComponent.instance.onSelection = (item) => {
+            // @ts-expect-error: As this point suggestion$ will be valid in the state.
             range.to = range.to + editor.view.state.suggestion$.query.length;
             execCommand({ editor: editor, range: range, props: item });
         };
--- a/libs/block-editor/src/lib/extensions/actions-menu.extension.ts
+++ b/libs/block-editor/src/lib/extensions/actions-menu.extension.ts
@@ -18,6 +18,8 @@ import {
 } from './components/suggestions/suggestions.component';
 import { ActionButtonComponent } from './components/action-button/action-button.component';
 import { PluginKey } from 'prosemirror-state';
+import { Subject } from 'rxjs';
+import { takeUntil } from 'rxjs/operators';
 
 declare module '@tiptap/core' {
     interface Commands<ReturnType> {
@@ -108,6 +110,7 @@ export const ActionsMenu = (viewContainerRef: ViewContainerRef) => {
     let myTippy;
     let suggestionsComponent: ComponentRef<SuggestionsComponent>;
     const suggestionKey = new PluginKey('suggestionPlugin');
+    const destroy$: Subject<boolean> = new Subject<boolean>();
 
     /**
      * Get's called on button click or suggestion char
@@ -117,13 +120,14 @@ export const ActionsMenu = (viewContainerRef: ViewContainerRef) => {
     function onStart({ editor, range, clientRect }: SuggestionProps | FloatingActionsProps): void {
         suggestionsComponent = getSuggestionComponent(viewContainerRef);
         suggestionsComponent.instance.onSelection = (item) => {
-            range.to = range.to + suggestionKey.getState(editor.view.state).query.length;
+            const suggestionQuery = suggestionKey.getState(editor.view.state).query?.length || 0;
+            range.to = range.to + suggestionQuery;
             execCommand({ editor: editor, range: range, props: item });
         };
-        suggestionsComponent.instance.clearFilter.subscribe(() => {
+        suggestionsComponent.instance.clearFilter.pipe(takeUntil(destroy$)).subscribe((type) => {
             const queryRange = {
                 to: range.to + suggestionKey.getState(editor.view.state).query.length,
-                from: range.from + 1
+                from: type === 'contentlet' ? range.from + 1 : range.from
             };
             editor.chain().deleteRange(queryRange).run();
         });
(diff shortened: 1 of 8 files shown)


libs/block-editor/src/lib/extensions/components/suggestions/suggestions.component.html
```diff
@@ -2,17 +2,19 @@
...
-            (mousedown)="onMouseDown($event, item)"
-            (mouseenter)="onMouseEnter($event)"
-            [command]="item.command"
-            [index]="i"
-            [label]="item.label"
-            [url]="item.icon"
-            [data]="item.data"
-        >
-        </dotcms-suggestions-list-item>
+        <ng-container *ngFor="let item of items; let i = index">
```
fmontes, line 5: Why?

libs/block-editor/src/lib/extensions/components/suggestions/suggestions.component.ts
```diff
@@ -196,6 +197,21 @@ export class SuggestionsComponent implements OnInit, AfterViewInit {
...
+     *
+     * @param {string} filter
+     * @memberof SuggestionsComponent
+     */
+    filterItems(filter: string = '') {
+        this.items.forEach((item) => {
+            item.visible = filter
+                ? item.label.toLowerCase().includes(filter.trim().toLowerCase())
+                : true;
+        });
```
fmontes, line 211: Why this `visible` instead of create a new array with the results? also, why not create a new array? that's the best practice, never mutate he original.

libs/block-editor/src/lib/extensions/dragHandler.extension.ts
```diff
@@ -132,7 +132,8 @@ export const DragHandler = (viewContainerRef: ViewContainerRef) => {
                                     nodeToBeDragged = getDirectChild(view.nodeDOM(position.inside));
                                     if (
                                         nodeToBeDragged &&
-                                        !nodeToBeDragged.classList?.contains('ProseMirror')
+                                        !nodeToBeDragged.classList?.contains('ProseMirror') &&
+                                        !nodeToBeDragged.innerText.startsWith('/')
```
fmontes, line 136: Move this to a method with a clear name, this will be getting bigger and bigger.

fmontes (changes requested):

libs/block-editor/src/lib/extensions/actions-menu.extension.ts
```diff
@@ -115,6 +115,8 @@ export const ActionsMenu = (viewContainerRef: ViewContainerRef) => {
     function onStart({ editor, range, clientRect }: SuggestionProps | FloatingActionsProps): void {
         suggestionsComponent = getSuggestionComponent(viewContainerRef);
         suggestionsComponent.instance.onSelection = (item) => {
+            // @ts-expect-error: As this point suggestion$ will be valid in the state.
+            range.to = range.to + editor.view.state.suggestion$.query.length;
```
rjvelazco, line 119: Here we can make a refactor to avoid the error:

`range.to = range.to + suggestionKey.getState(editor.view.state).query.length;`

In order to be able to use, we have to:

```js
const suggestionKey = new PluginKey('sugesstionPlugin');
....
    return Extension.create<FloatingMenuOptions>({
        name: 'actionsMenu',
        defaultOptions: {
            pluginKey: 'actionsMenu',
            suggestion: {
                char: '/',
                // Pass a pluginKey
                pluginKey: suggestionKey,
                allowSpaces: true,
                startOfLine: true,
             }
  ....
```

libs/block-editor/src/lib/extensions/actions-menu.extension.ts
```diff
@@ -170,14 +171,22 @@ export const ActionsMenu = (viewContainerRef: ViewContainerRef) => {
...
+                allowSpaces: true,
                 startOfLine: true,
                 render: () => {
                     return {
                         onStart,
                         onKeyDown,
                         onExit
                     };
+                },
+                items: ({ query }) => {
```
rjvelazco, line 183: Can we add a comment to the docs for future reference?

docs: https://tiptap.dev/api/utilities/suggestion#items

rjvelazco (changes requested):

libs/block-editor/src/lib/extensions/components/suggestions/suggestions.component.html
hmoreras: not needed anymore, but, two structural directives are not supported in the same element.

libs/block-editor/src/lib/extensions/components/suggestions/suggestions.component.ts
```diff
@@ -196,6 +197,21 @@ export class SuggestionsComponent implements OnInit, AfterViewInit {
...
+     *
+     * @param {string} filter
+     * @memberof SuggestionsComponent
+     */
+    filterItems(filter: string = '') {
+        this.items = [
+            ...this.initialItems.filter((item) =>
+                item.label.toLowerCase().includes(filter.trim().toLowerCase())
+            )
+        ];
```
fmontes, line 211: You don't need to do this, by default `filter` returns a new array.

fmontes (changes requested):

rjvelazco (approved):

libs/block-editor/src/lib/extensions/actions-menu.extension.ts
```diff
@@ -115,8 +117,16 @@ export const ActionsMenu = (viewContainerRef: ViewContainerRef) => {
...
+            range.to = range.to + suggestionKey.getState(editor.view.state).query.length;
             execCommand({ editor: editor, range: range, props: item });
         };
+        suggestionsComponent.instance.clearFilter.subscribe(() => {
+            const queryRange = {
+                to: range.to + suggestionKey.getState(editor.view.state).query.length,
+                from: range.from + 1
+            };
+            editor.chain().deleteRange(queryRange).run();
+        });
```
fmontes, line 129: Don't we need to unsubscribe here?

libs/block-editor/src/lib/extensions/actions-menu.extension.ts
fmontes: This

fmontes (changes requested):

libs/block-editor/src/lib/extensions/actions-menu.extension.ts
hmoreras: is done

fmontes (approved):

rjvelazco (approved):

PR #1969, opened 2022-05-02
dotCMS/core#22017 add delete block to bubble menu in images and contentlets

oidacra:
Add delete block to bubble menu in images and contentlets

Screen Shot 2022-05-02 at 3 32 54 PM

### DIFF
--- a/libs/block-editor/src/lib/plugins/dot-bubble-menu.plugin.ts
+++ b/libs/block-editor/src/lib/plugins/dot-bubble-menu.plugin.ts
@@ -23,7 +23,7 @@ import {
     isListNode,
     popperModifiers
 } from '../utils/bubble-menu.utils';
-import { findParentNode, getCountChildNodes, getNodeType } from '../utils/prosemirror.utils';
+import { findParentNode, getNodeType } from '../utils/prosemirror.utils';
 
 export const DotBubbleMenuPlugin = (options: DotBubbleMenuPluginProps) => {
     const component = options.component.instance;
@@ -272,6 +272,7 @@ export class DotBubbleMenuPluginView extends BubbleMenuView {
                 break;
             case 'deleteNode':
                 this.deleteSelectedNode();
+
                 break;
             case 'clearAll':
                 this.editor.commands.unsetAllMarks();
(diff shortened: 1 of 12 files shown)


libs/block-editor/src/lib/utils/prosemirror.utils.ts
```diff
@@ -0,0 +1,28 @@
...
+    let depth = selectionStart.depth;
+    let parent;
+    do {
+        parent = selectionStart.node(depth);
+        if (parent) {
+            if (Array.isArray(NodesTypesToFind)) {
+                if (NodesTypesToFind.includes(parent.type.name)) {
+                    break;
+                }
+            }
```
fmontes, line 22:
```suggestion
            if (Array.isArray(NodesTypesToFind) && NodesTypesToFind.includes(parent.type.name)) {
                break;
            }
```

fmontes (changes requested):

libs/block-editor/src/lib/utils/prosemirror.utils.ts
oidacra: Fixed.

libs/block-editor/src/lib/plugins/dot-bubble-menu.plugin.ts
```diff
@@ -373,4 +389,44 @@ export class DotBubbleMenuPluginView extends BubbleMenuView {
...
+            this.deleteSelectedCustomNodeType();
+        } else {
+            this.deleteSelectionNode();
+        }
+    }
+
+    private deleteSelectedCustomNodeType() {
+        const from = this.selection$FromPos.pos;
+        const to = from + 1;
+        this.editor.chain().deleteRange({ from, to }).blur().run();
```
rjvelazco, line 404: We should add a comment here because it's weird that the command does not work with `custom node`. It may be a problem with the command -[It's still a Draft](https://tiptap.dev/api/commands/delete-node)- but it can also be an issue with our `custom nodes`

The comment can be something like:
```
// TODO: Try to make the `deleteNode` command works with custom nodes.
```
or 

```
// TODO: Try the `deleteNode` command when a stable version is released
```

rjvelazco, line 404: We should add a comment here because it's weird that the command does not work with `custom node`. It may be a problem with the command -[It's still a Draft](https://tiptap.dev/api/commands/delete-node)- but it can also be an issue with our `custom nodes`

The comment can be something like:
```
// TODO: Try to make the `deleteNode` command works with custom nodes.
```
or 

```
// TODO: Try the `deleteNode` command when a stable version is released

```

libs/block-editor/src/lib/utils/prosemirror.utils.ts
```diff
@@ -0,0 +1,26 @@
...
+        if (parent) {
+            if (Array.isArray(NodesTypesToFind) && NodesTypesToFind.includes(parent.type.name)) {
+                break;
+            }
+            depth--;
+        }
+    } while (depth > 0 && parent);
+
+    return parent;
+};
```
rjvelazco, line 26: Nice!

rjvelazco, line 26: Nice!

rjvelazco (changes requested):

libs/block-editor/src/lib/plugins/dot-bubble-menu.plugin.ts
oidacra: Added

libs/block-editor/src/lib/plugins/dot-bubble-menu.plugin.ts
```diff
@@ -373,4 +389,44 @@ export class DotBubbleMenuPluginView extends BubbleMenuView {
...
+    }
+
+    private deleteSelectionNode() {
+        const selectionParentNode = findParentNode(this.selection$FromPos);
+        const nodeSelectionNodeType: NodeTypes = selectionParentNode.type.name;
+
+        switch (nodeSelectionNodeType) {
+            case NodeTypes.ORDERED_LIST:
```
hmoreras, line 412: this do something ?

fmontes (commented):
And also we have to point this to `release-22.06`

PR #1971, opened 2022-05-04
dotCMS/core#22044 Filter content types in suggestions 

hmoreras:
### Proposed Changes
* Allow Content Types filtering with a limit of 40 results.

### DIFF
--- a/libs/block-editor/src/lib/extensions/services/suggestions/suggestions.service.ts
+++ b/libs/block-editor/src/lib/extensions/services/suggestions/suggestions.service.ts
@@ -19,7 +19,7 @@ export class SuggestionsService {
 
     getContentTypes(filter = ''): Observable<DotCMSContentType[]> {
         return this.http
-            .get(`/api/v1/contenttype?filter=${filter}&orderby=name&direction=ASC`, {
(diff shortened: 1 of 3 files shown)


alfredo-dotcms (approved):

rjvelazco (approved):

libs/block-editor/src/lib/extensions/services/suggestions/suggestions.service.ts
```diff
@@ -19,7 +19,7 @@ export class SuggestionsService {
 
     getContentTypes(filter = ''): Observable<DotCMSContentType[]> {
         return this.http
-            .get(`/api/v1/contenttype?filter=${filter}&orderby=name&direction=ASC`, {
+            .get(`/api/v1/contenttype?filter=${filter}&orderby=name&direction=ASC&per_page=40`, {
```
fmontes, line 22: I would need to confirm with @wezell about this one because I think the 40 only applies is there is a filter applied.

fmontes: Confirmed, 40 and force to filter more.

oidacra:
Tested.

PR #1975, opened 2022-05-06
dotCMS/core#22039 Control what contentlet types you can add

hmoreras:
### Proposed Changes
* limit the type of content they can have in a block field.

### DIFF
--- a/apps/dotcms-block-editor/src/app/dot-block-editor/dot-block-editor.component.ts
+++ b/apps/dotcms-block-editor/src/app/dot-block-editor/dot-block-editor.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, Injector } from '@angular/core';
+import { Component, OnInit, Injector, Input } from '@angular/core';
 import { Editor } from '@tiptap/core';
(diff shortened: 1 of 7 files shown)


apps/dotcms-block-editor/src/app/dot-block-editor/dot-block-editor.component.ts
```diff
@@ -33,7 +33,7 @@ function toTitleCase(str) {
 })
 export class DotBlockEditorComponent implements OnInit {
     editor: Editor;
-
+    @Input() allowedContentTypes = '';
```
fmontes, line 36: Do we need this if we're getting this from the HTML?

fmontes (changes requested):

rjvelazco (approved):

apps/dotcms-block-editor/src/app/dot-block-editor/dot-block-editor.component.ts
hmoreras: yes, was changed here as input.

fmontes (approved):

